PR #2, opened 2018-04-18
using SIGSTOP (avoid interactive signal blocking)

amosbird:


amosbird:
`SIGTSTP` only works for processes attached to a tty.  I'm not sure what term should be used.

MaskRay:
`SIGSTOP` requires multiple `continue` commands in gdb while `SIGTSTP` only requires 1

amosbird:
https://github.com/scottt/debugbreak looks better

MaskRay:
I have changed the code to:
```
  const char* traceme = getenv("CCLS_TRACEME");
  if (traceme)
    raise(traceme[0] == 's' ? SIGSTOP : SIGTSTP);
```

`SIGTRAP` (debugbreak) only works when the process has been attached by a debugger, not in our use case.

MaskRay:
LLVM also supports `__builtin_debugtrap` (`llvm.debugtrap`) for this purpose.

amosbird:
Oh, OK. Good to know :)

PR #5, opened 2018-05-07
Option to let clang use libc++.

scturtle:
Make it an option.


scturtle:
I use ubuntu 16.04 with gcc 5.4 at work. `optional` can only be found as `experimental/optional`.

MaskRay:
I think this option is useful on Linux with older GCC (5.*)

The following snippet has been tested on FreeBSD 12.0-CURRENT and Mac OS X.
```
// line 99
if(${CMAKE_SYSTEM_NAME} STREQUAL Linux)
  target_link_libraries(ccls PRIVATE -lstdc++fs)
elseif(MSVC)
else()
  # e.g. Darwin, FreeBSD
  target_link_libraries(ccls PRIVATE -lc++experimental)
endif()
```

Is it possible to merge it with CLANG_USE_LIBC++?

MaskRay:
I have tested this option on Arch Linux.

With `archlinuxcn/{libc++,libc++abi,libc++experimental}`, `-DCLANG_USE_LIBC++=on` compiles
w/o, headers cannot be found.

scturtle:
Moved changes into `NOT SYSTEM_CLANG`...

scturtle:
I see. I always use the bundled clang to compile.

MaskRay:
I mean to use libc++ in the bundled clang+llvm downloaded from releases.llvm.org,
you need `-l` link options as well as `-isystem` include paths. Specifying the former only works because you have `libc++` `libc++abi` installed on your system.

I think these changes are still useful, for Linux distributions other than Arch.

PR #6, opened 2018-05-08
Random changes.

scturtle:
Refresh semantic highlight for all working files after loading the project.

### DIFF
--- a/src/clang_indexer.cc
+++ b/src/clang_indexer.cc
@@ -1841,12 +1841,15 @@ void OnIndexReference(CXClientData client_data, const CXIdxEntityRefInfo* ref) {
       if (!ns.def.spell) {
         ClangCursor sem_parent = referenced.get_semantic_parent();
         ClangCursor lex_parent = referenced.get_lexical_parent();
-        ns.def.spell =
-            SetUse(db, referenced.get_spell(), sem_parent, Role::Definition);
-        ns.def.extent =
-            SetUse(db, referenced.get_extent(), lex_parent, Role::None);
-        std::string name = referenced.get_spell_name();
-        SetTypeName(ns, referenced, nullptr, name.c_str(), param);
+        CXFile referenced_file;
+        Range spell = referenced.get_spell(&referenced_file);
+        if (file == referenced_file) {
+          ns.def.spell = SetUse(db, spell, sem_parent, Role::Definition);
+          ns.def.extent =
+              SetUse(db, referenced.get_extent(), lex_parent, Role::None);
+          std::string name = referenced.get_spell_name();
+          SetTypeName(ns, referenced, nullptr, name.c_str(), param);
+        }
       }
       break;
     }
--- a/src/import_pipeline.cc
+++ b/src/import_pipeline.cc
@@ -71,6 +71,13 @@ bool Indexer_Parse(DiagnosticsEngine* diag_engine,
   auto& request = *opt_request;
   ICacheManager cache;
 
+  // dummy one to trigger refresh semantic highlight
+  if (request.path.empty()) {
+    queue->on_indexed.PushBack(
+        Index_OnIndexed(IndexUpdate{}, PerformanceImportFile()), false);
+    return false;
+  }
+
   Project::Entry entry;
   {
     std::lock_guard<std::mutex> lock(project->mutex_);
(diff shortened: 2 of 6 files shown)


src/import_pipeline.cc
```diff
@@ -71,6 +71,14 @@ bool Indexer_Parse(DiagnosticsEngine* diag_engine,
   auto& request = *opt_request;
   ICacheManager cache;
 
+  if (request.path.empty()) { // dummy
+    IndexUpdate dummy;
+    dummy.file_id = -1;  // trigger refresh semantic highlight
+    queue->on_indexed.PushBack(
+        Index_OnIndexed(std::move(dummy), PerformanceImportFile()), false);
```
MaskRay: How about?

```c++
struct IndexUpdate {
...
  int file_id = -1;
...
```

`Index_OnIndexed(IndexUpdate{}, PerformanceImportFile())`

src/import_pipeline.cc
```diff
@@ -216,6 +224,18 @@ void QueryDb_OnIndexed(QueueManager* queue,
                        SemanticHighlightSymbolCache* semantic_cache,
                        WorkingFiles* working_files,
                        Index_OnIndexed* response) {
+
+  if (response->update.file_id < 0) { // dummy
```
MaskRay: Semantic highlighting does work reliably when the project is still loading is because `message_handler.cc` uses:

```c++
    switch (sym.kind) {
      case SymbolKind::Func: {
        const QueryFunc& func = db->GetFunc(sym);
        const QueryFunc::Def* def = func.AnyDef(); ///////// definition required, declaration is ignored
        if (!def)
          continue;  // applies to for loop
```

I think this patch is very promising to improve the situation! I occasionally observe that some files are not highlighted because I am unable to narrow it down (does not look like a forward-line out-of-range error)

I'll try this patch tomorrow on LLVM

src/include_complete.cc
```diff
@@ -106,8 +106,8 @@ void IncludeComplete::Rescan() {
     SetThreadName("scan_includes");
     Timer timer;
 
-    InsertIncludesFromDirectory(g_config->projectRoot,
-                                false /*use_angle_brackets*/);
+    // InsertIncludesFromDirectory(g_config->projectRoot,
```
MaskRay: This can just be removed. (I didn't even notice this because I don't use include completion much)

https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html

src/messages/initialize.cc
```diff
@@ -510,12 +510,9 @@ struct Handler_Initialize : BaseMessageHandler<In_InitializeRequest> {
       // Start indexer threads. Start this after loading the project, as that
       // may take a long time. Indexer threads will emit status/progress
       // reports.
-      if (g_config->index.threads == 0) {
-        // If the user has not specified how many indexers to run, try to
-        // guess an appropriate value. Default to 80% utilization.
-        g_config->index.threads =
-            std::max(int(std::thread::hardware_concurrency() * 0.8), 1);
-      }
+      if (g_config->index.threads == 0)
```
MaskRay: This is great!

src/project.cc
```diff
@@ -563,6 +563,9 @@ void Project::Index(QueueManager* queue,
     queue->index_request.PushBack(Index_Request(entry.filename, entry.args,
                                                 is_interactive, *content, id));
   });
+  // dummy request to indicate that project is loaded and
```
MaskRay: Nit. Capitalize.

src/clang_indexer.cc
```diff
@@ -1841,12 +1841,15 @@ void OnIndexReference(CXClientData client_data, const CXIdxEntityRefInfo* ref) {
...
         ClangCursor lex_parent = referenced.get_lexical_parent();
-        ns.def.spell =
-            SetUse(db, referenced.get_spell(), sem_parent, Role::Definition);
-        ns.def.extent =
-            SetUse(db, referenced.get_extent(), lex_parent, Role::None);
-        std::string name = referenced.get_spell_name();
-        SetTypeName(ns, referenced, nullptr, name.c_str(), param);
+        CXFile referenced_file;
+        Range spell = referenced.get_spell(&referenced_file);
+        if (file == referenced_file) {
```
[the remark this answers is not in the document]
scturtle: I think this should fix wrong spell name highlight for aliased namespace. For example, `namespace fs = std::experimental::filesystem` in `filesystem.hh`. `7:11-7:13` are also highlighted in other files.

MaskRay:
刪除增量更新一是對於大部分檔案沒什麼用(uses中行列號都要變化，只對尾部的修改有一點小作用，減輕主執行緒負擔；索引執行緒要寫索引到磁碟，沒有優化可言)；二是對於檔案頻繁保存的使用場景，不完全更新更容易產生引用重複/丟失問題。

我把timestamp_manager刪掉也是爲了簡化考慮(clang內部必然大量檔案系統stat，libclang下游節省一點stat沒太大意義)

PR #7, opened 2018-05-08
bundled libc++

scturtle:
Old linux guys can just use the bundled clang and libc++ to compile.


MaskRay:
Tested with:

```
cmake -G Ninja -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_EXE_LINKER_FLAGS=-fuse-ld=lld -DCMAKE_BUILD_TYPE=MinSizeRel -DCMAKE_EXPORT_COMPILE_COMMANDS=On -DCLANG_USE_BUNDLED_LIBC++=on -H. -Btest
```

Thanks!

PR #8, opened 2018-05-08
Fix file_id.

scturtle:
`int file_id = -1` doesn't works. It's set and used after this check.


MaskRay:
删除id remap后，file_id只有主线程知道了。昨天我忘记了这个

Issue #10, opened 2018-05-08
clang complete session assert error


scturtle:
An assert often fails in debug version:
```
    frame #5: 0x00000000007889f0 ccls`loguru::log_and_abort(stack_trace_skip=0, expr="CHECK FAILED:  !completion_sessions_.TryGet(filename)  ", file="../src/clang_complete.cc", line=774) at loguru.hpp:2661
    frame #6: 0x00000000004ec504 ccls`ClangCompleteManager::TryGetSession(this=0x00007ffda03b6450, filename="XXX.cc", mark_as_completion=true, create_if_needed=true) at clang_complete.cc:774
    frame #7: 0x00000000004f99b4 ccls`(anonymous namespace)::DiagnosticQueryMain(completion_manager=0x00007ffda03b6450) at clang_complete.cc:593
    frame #8: 0x00000000004f9910 ccls`ClangCompleteManager::ClangCompleteManager(this=0x00000000028532c8)>, std::__1::function<void (lsRequestId)>)::$_2::operator()() const at clang_complete.cc:667
```
Not know the meaning. Dig later.

MaskRay:
I think completion is also unnecessarily complex but I haven't looked into how clangd does this yet.

MaskRay:
Removed the `assert`

PR #12, opened 2018-05-14
Bring back old completion features.

scturtle:


src/messages/text_document_completion.cc
```diff
@@ -365,7 +365,7 @@ struct Handler_TextDocumentCompletion : MessageHandler {
             lock.lock();
           std::string quote = result.match[5];
           for (auto& item : include_complete->completion_items)
-            if (quote.empty() || quote == (item.use_angle_brackets_ ? "<" : "\""))
+            // if (quote.empty() || quote == (item.use_angle_brackets_ ? "<" : "\""))
```
MaskRay: I don't remember what this is about.

`-I` available in both `#include "` and `#include <`
`-isystem` available in `#include <`
`-iquote` available in `#include "`

src/standard_headers.inc
```diff
@@ -0,0 +1 @@
...rrno", "cfenv", "cfloat", "chrono", "cinttypes", "ciso646", "climits", "clocale", "cmath", "codecvt", "complex", "complex.h", "condition_variable", "coroutine", "csetjmp", "csignal", "cstdarg", "cstdbool", "cstddef", "cstdint", "cstdio", "cstdlib", "cstring", "ctgmath", "ctime", "ctype.h", "cwchar", "cwctype", "cxxabi.h", "deque", "dynarray", "errno.h", "exception", "experimental", "ext", "filesystem", "float.h", "forward_list", "fstream", "functional", "future", "initializer_list", "inttypes.h", "iomanip", "ios", "iosfwd", "iostream", "istream", "iterator", "limits", "limits.h", "list", "locale", "locale.h", "map", "math.h", "memory", "memory_resource", "module.modulemap", "mutex", "new", "numeric", "optional", "ostream", "propagate_const", "queue", "random", "ratio", "regex", "scoped_allocator", "set", "setjmp.h", "shared_mutex", "sstream", "stack", "stdbool.h", "stddef.h", "stdexcept", "stdint.h", "stdio.h", "stdlib.h", "streambuf", "string", "string.h", "string_view", "strstream", "support", "system_error", "tgmath.h", "thread", "tuple", "typeindex", "typeinfo", "type_traits", "unordered_map", "unordered_set", "utility", "valarray", "variant", "vector", "wchar.h", "wctype.h" };
```
MaskRay: This list made me sad. I think the header completion can be achieved with either

```emacs
;; doom-emacs
  (set! :company-backend '(c-mode c++-mode) '(company-lsp company-c-headers))
```

or

```emacs
(setq ccls-extra-init-params '(:clang (:extraArgs ["-cxx-isystem", "/usr/include/c++/v1", "-c-isystem", "/usr/include"])))
```

MaskRay: I also tend to avoid duplicating cquery's logic of "detecting system include paths with an external program"

src/clang_complete.cc
```diff
@@ -776,7 +776,7 @@ std::shared_ptr<CompletionSession> ClangCompleteManager::TryGetSession(
     // If this request is for a completion, we should move it to
     // |completion_sessions|.
     if (mark_as_completion) {
-      assert(!completion_sessions_.TryGet(filename));
+      // assert(!completion_sessions_.TryGet(filename));
```
MaskRay: Related to https://github.com/MaskRay/ccls/issues/10 ?

src/messages/text_document_completion.cc
scturtle: I see. `-I` now adds dir to both categories.

src/standard_headers.inc
scturtle: The latter solution seems to be OK. It's handy to provide basic system header completion out-of-box.

src/clang_complete.cc
scturtle: Yeah. Seems to be not harmful.

PR #31, opened 2018-07-14
Fix cmake/FindClang.cmake on FreeBSD

khng300:
execinfo is not specified as linking dependencies, but libLLVMSupport requires it.
Specify execinfo as linking dependency for FreeBSD in cmake/FindClang.cmake.


MaskRay:
commit 0a304096df741355ccaa4ab1983502783dd96768 removed `Backtrace` (which I think means `-lexecinfo` on FreeBSD). I have applied the change in `CMakeLists.txt`, instead of `cmake/FindClang.cmake`.

PR #35, opened 2018-07-25
Fix preload completion session.

scturtle:


src/match.cc
```diff
@@ -25,7 +25,7 @@ std::optional<Matcher> Matcher::Create(const std::string& search) {
         // std::regex_constants::nosubs
     );
     return m;
-  } catch (std::exception e) {
+  } catch (const std::exception& e) {
```
MaskRay: Nice catch. Otherwise copy initialization happens.

PR #36, opened 2018-07-27
Fix diagnostics on MacOS and duplicated bases.

scturtle:


src/indexer.cc
```diff
@@ -1222,6 +1222,7 @@ std::vector<std::unique_ptr<IndexFile>> Index(
       Uniquify(it.second.uses);
       // e.g. declaration + out-of-line definition
       Uniquify(it.second.def.funcs);
+      Uniquify(it.second.def.bases);
```
MaskRay: swap this line with `funcs`?

scturtle:
Yes for MacOS but no for linux. Even specified in `LoadFromCompilerInvocationAction` of `indexer.cc`.

MaskRay:
Do you mean:

* For Mac OS X, `g_config->clang.resourceDir` needs to be passed to `ASTUnit::createFromCommandLine`
* For Linux, `-resource-dir` may be explicitly passed

?

scturtle:
Yes.

MaskRay:
Thanks!

PR #44, opened 2018-08-03
add flat option to ccls/inheritanceHierarchy 

amosbird:
useful for xref-like examination.

### DIFF
--- a/src/messages/ccls_inheritanceHierarchy.cc
+++ b/src/messages/ccls_inheritanceHierarchy.cc
@@ -1,6 +1,8 @@
 #include "message_handler.h"
 #include "query_utils.h"
 #include "pipeline.hh"
+#include "log.hh"
+
 using namespace ccls;
 
 #include <unordered_set>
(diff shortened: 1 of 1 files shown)


MaskRay:
what's the difference between this and $ccls/inheritanceHierarchy ?

amosbird:
It's used to export `xrefs` via `(lsp--locations-to-xref-items)` like commands `ccls/callers, ccls/vars` etc

MaskRay:
Which do you think it resembles more, `$ccls/inheritanceHierarchy` or `$ccls/{base,derived}`? I wonder if it is possible to unify it with either existing interface, instead of creating (duplicating) a new one.

amosbird:
`$ccls/{base,derived}`  . Well it's possible to traverse the entry-tree in elisp code but I feel that using a dedicated commands for that is much simpler and also provides support for other clients which doesn't have hierarchy interfaces.

MaskRay:
I agree, that is why I asked LanguageClient-neovim author to support custom xref long ago https://github.com/autozimu/LanguageClient-neovim/issues/454 :)

I think this file should be unified with `$ccls/inheritanceHierarchy`.

amosbird:
OK, is it fine to use `lsLocation` instead of `lsLocationEx` for xrefs?

MaskRay:
The `flag` option has been added

PR #46, opened 2018-08-05
Fix compile on Windows + MSYS2-MinGW64 + GCC8

Riatre:
1. libClangDriver.a shipped by MSYS2 needs -lversion which is not
   included in CMAKE_CXX_STANDARD_LIBRARIES by default.
2. Use string literal for DEFAULT_RESOURCE_DIRECTORY to support
   backslashes in path.


MaskRay:
Will you come to Las Vegas?

Riatre:
Likely.

PR #61, opened 2018-08-29
fix clang_th.h && clang_complete.cc

FirstLoveLife:
fix compile errors for your latest commit.
```
clang version 6.0.1 (tags/RELEASE_601/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```


MaskRay:
thx! i made some adjustments and merged

PR #73, opened 2018-09-11
Normalize paths on Windows

Riatre:
Generally, we have three sources of path strings:

1. LSP document URI and project root.
2. Clang indexer.
3. compile_commands.json or GetFileFromDirectory.

Path strings from these three sources may have different slash (forward/backward) and drive letter (lowercase/uppercase) styles, as we use file path as map key, we have to normalize them.

Slashes will be converted to forward slash (`'/'`), drive letters will be converted to uppercase (On Windows, it is guaranteed to be case insensitive).


Riatre:
Note that this does not address https://github.com/MaskRay/emacs-ccls/issues/10#issuecomment-419799769.
It is caused by a different issue: the auto detection logic in `lib/Tooling/JSONCompilationDatabase.cpp` thinks the command line is in GNU syntax (where '\\' is a escape character instead of path separator) so it eats all '\\' in paths. I'll try to fix this in another PR.

PR #78, opened 2018-09-22
Remove chunks with CK_Information kind from insertText.

### LINKED ISSUE #77: Wrong suggestion with const member function.
2018-09-22-132919_1920x1080_scrot

I'm using ccls with neovim and LanguageClient-neovim. ccls was intalled from Arch Linux repo `archlinuxcn`

Riatre:
Without this ccls inserts "size() const" in the following scenario:

std::string text;
text.si| <-- Trigger completion here and pick "size"

Fixes #77


MaskRay:
Thanks!

PR #79, opened 2018-09-22
 Allow force disabling snippet via initializationOptions

Riatre:


src/config.h
```diff
@@ -84,7 +84,7 @@ struct Config {
     // TextDocumentClientCapabilities.documentSymbol.hierarchicalDocumentSymbolSupport
     bool hierarchicalDocumentSymbolSupport = false;
     // TextDocumentClientCapabilities.completion.completionItem.snippetSupport
-    bool snippetSupport = false;
+    bool snippetSupport = true;
```
MaskRay: Did you find an editor which supports LSP-style snippets but does not announce it?

src/messages/textDocument_completion.cc
```diff
@@ -634,6 +634,7 @@ struct Handler_TextDocumentCompletion
 #if LLVM_VERSION_MAJOR >= 7
       CCOpts.IncludeFixIts = true;
 #endif
+      CCOpts.IncludeMacros = true;
```
MaskRay: Thanks! cherry-picked this commit

src/messages/textDocument_documentSymbol.cc
```diff
@@ -204,6 +204,12 @@ struct Handler_TextDocumentDocumentSymbol
             if (!def || !def->spell || def->is_local())
               continue;
           }
+          if (!params.all && sym.kind == SymbolKind::Type) {
```
MaskRay: Do you use hierarchical document symbols?

src/clang_complete.cc
```diff
@@ -165,10 +167,11 @@ std::unique_ptr<CompilerInstance> BuildCompilerInstance(
     CompletionSession &session, std::unique_ptr<CompilerInvocation> CI,
     DiagnosticConsumer &DC, const WorkingFiles::Snapshot &snapshot,
     std::vector<std::unique_ptr<llvm::MemoryBuffer>> &Bufs) {
-  StringRef main = CI->getFrontendOpts().Inputs[0].getFile();
+  auto main =
```
MaskRay: CI->getFrontendOpts().Inputs[0].getFile() can be `"../src/stdio/asprintf.c"` because `compile_commands.json` main source file is not normalized now.

MaskRay: Regarding 
"Add path normalization around clang::FrontendInputFile::getFile()"

I moved `ResolveIfRelative` to `utils.h` and use that to address the issue.

src/config.h
Riatre: No. This is for force **disable** the snippet support instead of force enable.

Before this change `snippetSupport` is override by `params.capabilities.completion.completionItem.snippetSupport` in `initialize`.
I believe it makes sense to enable snippet support only when `params.capabilities.completion.completionItem.snippetSupport == true && client.snippetSupport == true` as some clients does not support configuring the capabilities it announce.

src/messages/textDocument_documentSymbol.cc
Riatre: Thanks for implementing a better fix!

src/config.h
MaskRay: OK. I think you mean even if the client announces that it supports snippets, i.e. `params.capabilities.completion.completionItem.snippetSupport == true`

Users should be able to disable it with `client.snippetSupport: false`

I think this change also applies to `hierarchicalDocumentSymbolSupport`

PR #82, opened 2018-09-26
Update threaded_queue.h

amosbird:
https://en.cppreference.com/w/cpp/language/fold

### DIFF
--- a/src/threaded_queue.h
+++ b/src/threaded_queue.h
@@ -48,8 +48,7 @@ private:
   }
 
   template <size_t... Is> void unlock_impl(std::index_sequence<Is...>) {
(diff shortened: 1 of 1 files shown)


src/threaded_queue.h
```diff
@@ -48,8 +48,7 @@ template <typename... Queue> struct MultiQueueLock {
   }
 
   template <size_t... Is> void unlock_impl(std::index_sequence<Is...>) {
-    (void)std::initializer_list<int>{
```
MaskRay: In C++11 or C++14, MSVC had a bug and this construction was required to work around it. Since we are now using C++17, this is no longer relevant.

PR #84, opened 2018-09-30
Rename some initialization options

MaskRay:
* Delete index.enabled which can be achieved with index.blacklist: ['.']
* Move completion.include* to completion.include.*
* move largeFileSize to highlight.largeFileSize


MaskRay:
@Riatre @scturtle

MaskRay:
Merged

PR #86, opened 2018-10-02
 regard conversion as method instead of constructor

FirstLoveLife:
1. According to http://eel.is/c++draft/class.ctor and http://eel.is/c++draft/#class, user-defined implicit conversion should be regarded as the constructor. 
2. 

```
case Decl::CXXConstructor:	   
    case Decl::CXXConversion:	
      func->def.kind = lsSymbolKind::Constructor;
```

will be weird in lsp-imenu:
```
(Class)
  ┃ operator bool (Constructor)
  ┃ operator& (Method)
  ┃ __bit_const_reference<_Cp> (Constructor)
  ┃ __bit_const_reference<_Cp> (Constructor)
  ┃ __storage_pointer (Other)
  ┃ __storage_type (Other)
  ...
```

Clearly, operator bool is not this class's `Constructor`. This commit changes the layout into:

```
(Class)
  ┃ operator bool (Method)
  ┃ operator& (Method)
  ┃ __bit_const_reference<_Cp> (Constructor)
  ┃ __bit_const_reference<_Cp> (Constructor)
  ┃ __storage_pointer (Other)
  ┃ __storage_type (Other)
  ...
```


src/indexer.cc
```diff
@@ -1033,10 +1033,10 @@ class IndexDataConsumer : public index::IndexDataConsumer {
       break;
     }
     case Decl::CXXConstructor:
-    case Decl::CXXConversion:
       func->def.kind = lsSymbolKind::Constructor;
       break;
     case Decl::CXXDestructor:
```
MaskRay: Can you swap them?

.gitignore
```diff
@@ -3,3 +3,4 @@ build
 debug
 release
 /compile_commands.json
+Release
```
MaskRay: It was a fault that I (or someone else) put debug/release in `.gitignore`

src/indexer.cc
FirstLoveLife: Ok

FirstLoveLife: @MaskRay Updated

MaskRay:
cherry picked "regard conversion as method instead of constructor". Thanks!

PR #89, opened 2018-10-09
Fix additionalTextEdits -> textEdit hack for VS Code

Riatre:
Visual Studio Code filters the completion result according to
textEdit.range and filterText, if the textEdit.range overlaps with
existing text, we have to include the overlapping texts in filterText,
otherwise the completion item would be filtered out.

See https://github.com/Microsoft/vscode/issues/41208, even their own code uses such hack: https://github.com/Microsoft/vscode/commit/5fe3f97edb64ce2b518c47123d281182d245b9f2

Alternatively, since clang only supports correcting "." to "->" and "->" to "." right now, we may simply check `edits[0].newText` and insert its opposite before `filterText`. Let me know if you prefer this.

### DIFF
--- a/src/messages/textDocument_completion.cc
+++ b/src/messages/textDocument_completion.cc
@@ -155,10 +155,20 @@ template <typename T> char *tofixedbase64(T input, char *out) {
 // when given 1000+ completion items.
 void FilterCandidates(Out_TextDocumentComplete *complete_response,
                       const std::string &complete_text, lsPosition begin_pos,
-                      lsPosition end_pos, bool has_open_paren,
-                      const std::string &buffer_line) {
+                      lsPosition end_pos, const std::string &buffer_line) {
+  assert(begin_pos.line == end_pos.line);
   auto &items = complete_response->result.items;
 
+  // People usually does not want to insert snippets or parenthesis when
+  // changing function or type names, e.g. "str.|()" or "std::|<int>".
+  bool has_open_paren = false;
+  for (int c = end_pos.character; c < buffer_line.size(); ++c) {
+    if (buffer_line[c] == '(' || buffer_line[c] == '<')
+      has_open_paren = true;
+    if (!isspace(buffer_line[c]))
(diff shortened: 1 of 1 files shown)


src/messages/textDocument_completion.cc
```diff
@@ -155,7 +155,8 @@ template <typename T> char *tofixedbase64(T input, char *out) {
 // when given 1000+ completion items.
 void FilterCandidates(Out_TextDocumentComplete *complete_response,
                       const std::string &complete_text, lsPosition begin_pos,
-                      lsPosition end_pos, bool has_open_paren) {
+                      lsPosition end_pos, bool has_open_paren,
+                      const std::string &buffer_line) {
```
MaskRay: If you are happy to do more...

Since `buffer_line` is now passed as a parameter, `has_open_paren` can be removed and the `IsOpenParenOrAngle` logic can be inlined here (it can be simplified a bit).

src/messages/textDocument_completion.cc
```diff
@@ -173,8 +174,16 @@ void FilterCandidates(Out_TextDocumentComplete *complete_response,
       // Order of textEdit and additionalTextEdits is unspecified.
       auto &edits = item.additionalTextEdits;
       if (edits.size() && edits[0].range.end == begin_pos) {
-        item.textEdit.range.start = edits[0].range.start;
+        auto&& start = edits[0].range.start;
```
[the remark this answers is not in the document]
Riatre: Done.

src/messages/textDocument_completion.cc
Riatre: Done.

This feature is actually already in a broken state in master:

1. `auto [c, l] = position;` but position has `int line; int character;`.
2. It inserts `item.label` which is wrong if detailedLabel = true.

MaskRay:
Thanks! And sorry for asking you to do the extra cleanup :)

PR #91, opened 2018-10-10
Fix hierarchical document symbol

Riatre:
1. Fixed a bug on building document symbol tree: As `sym2ds` was updated in place, nested funcs/types may be moved into `children` of another `lsDocumentSymbol` before itself got processed.
2. Namespaces only have declarations, in the old implementation it wasn't included in the result, making the result less hierarchical. This commit fixes this by including the declarations of a symbol if no definitions found.

### DIFF
--- a/src/messages/textDocument_documentSymbol.cc
+++ b/src/messages/textDocument_documentSymbol.cc
@@ -16,6 +16,7 @@ limitations under the License.
 #include "message_handler.h"
 #include "pipeline.hh"
 #include "query_utils.h"
+#include "log.hh"
 using namespace ccls;
 using namespace clang;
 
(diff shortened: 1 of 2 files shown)


src/messages/textDocument_documentSymbol.cc
```diff
@@ -112,92 +118,100 @@ struct Handler_TextDocumentDocumentSymbol
...
               else
-                break;
-              kv.first = static_cast<const void *>(&def);
+                continue;
+              def_ptr = &def;
+              break;
             }
+          
+          // Try to find a declaration.
+          if (!def_ptr && candidate_def_ptr)
```
MaskRay: Forward declarations have `entity.declarations` but no `entity.def.spell`. I think they are usually unwanted, shall we apply this heuristic only for `lsSymbolKind::Namespace`?

MaskRay:
Cherry-picked with small style change. For namespaces, include the declarations of a symbol if no definitions found.

PR #103, opened 2018-10-23
-DSYSTEM_CLANG=OFF: allow using Ubuntu 14.04

dfrib:
In the 6.0.1 -> 7.0.0 LLVM migration of f27639babfd7a29c8c76681e99c015d1f3e4a676, archive hashes were added for the LLVM 7.0.0 binaries not only for Ubuntu 16.04, but also for Ubuntu 14.04. Afaics, the `DownloadAndExtractClang.cmake` functionality would never allow making use of the latter, however; always using the Ubuntu 16.04 LLVM 7.0.0 binary in case of OS Linux.

This commit generalizes the archive name derivation for when the current platform is Ubuntu, while persisting using Ubuntu 16.04 as default archive in case CMAKE_SYSTEM_NAME is Linux (even if the actual dist is not Ubuntu).

This change will also facilitate keeping backwards compatibility with Ubuntu 16.04 (possibly 14.04) for the upcoming LLVM 8.0.0 migration, where an archive for Bionic (18.04) is expected to be included (which may be chosen as the default).

If this just adds unneeded complexity to `DownloadAndExtractClang.cmake`, we could alternatively simply remove `clang_archive_hashes/clang+llvm-7.0.0-x86_64-linux-gnu-ubuntu-14.04.tar.xz.SHA256`.

### DIFF
--- a/cmake/DownloadAndExtractClang.cmake
+++ b/cmake/DownloadAndExtractClang.cmake
@@ -12,8 +12,41 @@ set(CLANG_ARCHIVE_EXT .tar.xz)
 if(${CMAKE_SYSTEM_NAME} STREQUAL Linux)
 
   # Default to Ubuntu 16.04
-  set(CLANG_ARCHIVE_NAME
-      clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-16.04)
+  set(DEFAULT_UBUNTU_VERSION_NUMBER 16.04)
+
+  # Select a Clang archive for other Ubuntu releases if the current platform
+  # differs from the default one
+  set(UBUNTU_VERSION_NUMBER ${DEFAULT_UBUNTU_VERSION_NUMBER})
+  find_program(LSB_RELEASE_EXECUTABLE lsb_release)
+
+  if(NOT LSB_RELEASE_EXECUTABLE)
+    message(WARNING "Could not find lsb_release executable. \
+Falling back to default Linux platform: Clang/LLVM archive for \
+Ubuntu ${DEFAULT_UBUNTU_VERSION_NUMBER}.")
+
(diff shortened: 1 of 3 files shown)


MaskRay:
Thanks for contributing. I copyed the clang hashes from @DaanDeMeyer's cquery PR. (Thanks to all his work :) if you dig up ccls's history of `-DSYSTEM_CLANG=on` upgrade)

However, I'm not particularly in favor of this change (the additional 30+ lines of cmake complexity). Ubuntu 14.04 seems too old to me.

ccls has decided to enforce C++17 and [require GCC >= 7.3](https://github.com/MaskRay/ccls/wiki/Getting-started) (GCC 7.2 probably also works; or clang >= 5). Ubuntu 14.04 users need some work to make ccls build.

They can also read instructions on https://github.com/MaskRay/ccls/wiki/Build#cmake-options , download and unpack the prebuilt archive by themselves, then use:

`cmake -GNinja -H. -BBundled -DSYSTEM_CLANG=on -DCMAKE_PREFIX_PATH=$PWD/build/clang+llvm-7.0.0-x86_64-linux-gnu-ubuntu-14.04`

dfrib:
I would be happy to start contributing to this project, but I have a lot to learn about it (and Clang/LLVM) first :)

I see - the complexity did indeed grow a bit big considering the gain.

Should we remove the Ubuntu 14.04 hash, or leave it for 14.04 users doing custom local fixes in order to build `ccls`?

MaskRay:
Thanks :)

I believe 7.0.1 release will come soon. Maybe we can delete the Ubuntu 14.04 hash by then...

dfrib:
That sounds good. I’m closing this PR.

PR #104, opened 2018-10-24
Declare documentLink is not supported in capabilities

Riatre:
Otherwise clients may still send textDocument/documentLink requests and wait indefinitely for responses.


MaskRay:
Just implemented this...

Riatre:
Never mind then

PR #109, opened 2018-11-05
Announce openClose support in ServerCap.TextDocumentSyncOptions

Riatre:
For some mysterious reason my VSCode suddenly stops sending didOpen / didClose notifications, causing breakage in almost all features. I dug a bit and found that openClose was set to false (i.e. not supported) in ServerCap, this is obviously wrong, and after setting it to true everything goes back to normal.

I have absolutely no idea why it worked well before.

Also, there is a extra message_handler.hh outside `src/`, it is removed in this diff.


MaskRay:
Sorry, there was a comment saying that `TextDocumentSyncOptions` did not work. In the recent `initialize.cc` refactor, I made the `textDocumentSync: number` -> `textDocumentSync: TextDocumentSyncOptions` change. Thanks for identifying the issue!

```typescript
interface ServerCapabilities {
	/**
	 * Defines how text documents are synced. Is either a detailed structure defining each notification or
	 * for backwards compatibility the TextDocumentSyncKind number. If omitted it defaults to `TextDocumentSyncKind.None`.
	 */
	textDocumentSync?: TextDocumentSyncOptions | number;
```

PR #117, opened 2018-11-11
migrate CCLS to windows

DsoTsin:
Add appveyor ci status icon to ReadMe.md

### DIFF
--- a/src/clang_tu.cc
+++ b/src/clang_tu.cc
@@ -45,7 +45,8 @@ static Pos Decomposed2LineAndCol(const SourceManager &SM,
                                  std::pair<FileID, unsigned> I) {
   int l = SM.getLineNumber(I.first, I.second) - 1,
       c = SM.getColumnNumber(I.first, I.second) - 1;
-  return {(int16_t)std::min(l, INT16_MAX), (int16_t)std::min(c, INT16_MAX)};
+  return {(int16_t)std::min<int16_t>(l, INT16_MAX),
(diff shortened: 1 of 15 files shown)


src/clang_tu.cc
```diff
@@ -45,7 +45,8 @@ static Pos Decomposed2LineAndCol(const SourceManager &SM,
                                  std::pair<FileID, unsigned> I) {
   int l = SM.getLineNumber(I.first, I.second) - 1,
       c = SM.getColumnNumber(I.first, I.second) - 1;
-  return {(int16_t)std::min(l, INT16_MAX), (int16_t)std::min(c, INT16_MAX)};
+  return {(int16_t)std::min<int16_t>(l, INT16_MAX),
```
MaskRay: I am a bit surprised. Doesn't `(int16_t)std::min(l, INT16_MAX)` compile on MSVC 2017?

`std::min<int16_t>` narrow converts `l` to `int16_t`, which is not intended.

DsoTsin: It won't be compiled on VC2017, since the compiler failed to infer the parameter type by return type.

src/position.hh
```diff
@@ -76,13 +76,14 @@ void Reflect(Writer &visitor, Range &value);
 namespace std {
 template <> struct hash<ccls::Range> {
   std::size_t operator()(ccls::Range x) const {
-    union U {
+    /*union U {
```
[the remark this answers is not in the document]
DsoTsin: Ok, it works.

PR #118, opened 2018-11-12
Fix VS2017 Compile issues and Add CI script

DsoTsin:


src/position.hh
```diff
@@ -76,6 +76,7 @@ namespace std {
 template <> struct hash<ccls::Range> {
   std::size_t operator()(ccls::Range x) const {
     union U {
+      U() {} // On VS2017, still need constructor here
```
MaskRay:
```
    union U {
      U() {}
      ccls::Range range;
      uint64_t u64;
    } u;
```

MaskRay: No comment is needed

cmake/FindClang.cmake
```diff
@@ -143,9 +143,10 @@ if(Clang_FOUND AND NOT TARGET Clang::Clang)
   set_target_properties(Clang::Clang PROPERTIES
     IMPORTED_LOCATION ${Clang_LIBRARY}
     INTERFACE_INCLUDE_DIRECTORIES "${Clang_INCLUDE_DIR};${Clang_BUILD_INCLUDE_DIR};${LLVM_INCLUDE_DIR};${LLVM_BUILD_INCLUDE_DIR}")
-
-  find_package(Curses REQUIRED)
-  find_package(ZLIB REQUIRED)
+  if(NOT WIN32)
```
MaskRay: Does msys2 build need Curses & ZLIB ?

CMakeLists.txt
```diff
@@ -58,8 +58,11 @@ if(MSVC)
             # roughly -Wno-sign-compare
```
MaskRay: I think `/wd4996` `/wd4722` may be unnecessary as loguru has been deleted for a long time... Can you do me a favor to check if that is the case?

.appveyor.yml
```diff
@@ -0,0 +1,22 @@
...
+
+platform:
+  - x64
+
+build:
+  parallel: true  # enable MSBuild parallel builds
+  verbosity: minimal
+
+install:
+  - if not exist llvm.tar.xz appveyor DownloadFile "https://ziglang.org/deps/llvm+clang-7.0.0-win64-msvc-release.tar.xz" -FileName llvm.tar.xz
```
MaskRay: Question: is this the best you can find? Will it download the archive again and again?

cmake/FindClang.cmake
DsoTsin: I don't have msys2 environments. With VS2017 toolchain, these libs are not required.

CMakeLists.txt
DsoTsin: /wd4996 is still needed, 4772 warning does't appear under MSVC 2017.

DsoTsin: Your can check the logs here, https://ci.appveyor.com/project/DsoTsin/ccls/builds/20224202

cmake/FindClang.cmake
MaskRay: I've changed it to `if(NOT MSVC)`. I think msys2 llvm still uses it (but I don't have a Windows environment to test)

MaskRay:
Please check if my amended version still builds

src/position.hh
myrgy: You may remove constryctor if you will swapunion fields.

MaskRay: > You may remove constryctor if you will swapunion fields.

What's the rule applied :)

myrgy: In general thats because defaul definition is not trivially constructible. You may try to add constexpr or remove initialization from Pos or swap union members.
In that case union will construct uint instead of Range. Please see https://ideone.com/iiHf5q

PR #135, opened 2018-11-24
Make lsRange in documentLink narrower.

### LINKED ISSUE #132: Include incorrectly recognized as link.
A C++ include in code is incorrectly recognized as a link, and it offers "click to follow link" option on hover. See below:

screen shot 2018-11-23 at 10 43 09 pm

This should be disabled -- an include is not a link.

Riatre:
Fixes #132, kind of.

I just happen to have some naive "tokenization work" patch mentioned in https://github.com/MaskRay/ccls/issues/132#issuecomment-441293384 since without this, documentLink looks really ugly in VSCode.


MaskRay:
Thanks! I changed to a slightly compacter version.

PR #146, opened 2018-12-04
Track data flow into variables and function returns

LeszekSwirski:


src/indexer.cc
```diff
@@ -830,6 +831,125 @@ class IndexDataConsumer : public index::IndexDataConsumer {
            D->getKind() == Decl::CXXConversion))
         role = Role(role | Role::Implicit);
       do_def_decl(func);
+
+      if (is_def) {
+        class FuncVisitor : public RecursiveASTVisitor<FuncVisitor> {
+          public:
+            FuncVisitor(IndexDataConsumer* parent, 
+    IndexFile *db, SourceManager& SM, const LangOptions &Lang, int lid, const Decl* D) : parent(parent), db(db), SM(SM), Lang(Lang), lid(lid), root(D) {}
```
MaskRay: Misindented.

src/indexer.cc
```diff
@@ -830,6 +831,125 @@ class IndexDataConsumer : public index::IndexDataConsumer {
            D->getKind() == Decl::CXXConversion))
         role = Role(role | Role::Implicit);
       do_def_decl(func);
+
+      if (is_def) {
+        class FuncVisitor : public RecursiveASTVisitor<FuncVisitor> {
+          public:
+            FuncVisitor(IndexDataConsumer* parent, 
+    IndexFile *db, SourceManager& SM, const LangOptions &Lang, int lid, const Decl* D) : parent(parent), db(db), SM(SM), Lang(Lang), lid(lid), root(D) {} 
+            bool VisitFunctionDecl(FunctionDecl* decl) {
```
MaskRay: I've changed `.clang-format` to use the LLVM style.. so the argument should be written as `FunctionDecl *decl`. It may look weird as they have different CamelCase/camelCase conventions. I also find it a bit unnatural to fix existing Chromium style into LLVM but what I chose is that: for llvm related variables/types, use its naming convention (`VariableName` `funcName`) but for others use the Chromium one.

src/indexer.cc
```diff
@@ -830,6 +831,125 @@ class IndexDataConsumer : public index::IndexDataConsumer {
...
+              return true;
+            }
+            bool VisitBinAssign(BinaryOperator* binop) {
+              AddDataFlow(binop->getLHS(), binop->getRHS());
+              return RecursiveASTVisitor::VisitBinAssign(binop);
+            }
+            bool VisitVarDecl(VarDecl* decl) {
+              if (decl->hasInit()) {
```
MaskRay: These parentheses wrapping simple statements are very annoying. LLVM style tends to avoid them.

src/indexer.cc
```diff
@@ -830,6 +831,125 @@ class IndexDataConsumer : public index::IndexDataConsumer {
...
+                auto* data_flow_decl = data_flowExpr->getMemberDecl();
+                write.role = Role::Read;
+                return DataFlow{parent->GetUsr(data_flow_decl), write};
+              } else if (const auto* data_flowExpr = dyn_cast<CallExpr>(source->IgnoreParenCasts())) {
+                if (auto* data_flow_callee = data_flowExpr->getDirectCallee()) {
+                  write.role = Role::Call;
+                  return DataFlow{parent->GetUsr(data_flow_callee), write};
+                }
+              }
+
```
MaskRay: `^\s+$`

src/indexer.cc
```diff
@@ -830,6 +831,125 @@ class IndexDataConsumer : public index::IndexDataConsumer {
...
+              } else if (const auto* targetFieldRef = dyn_cast<MemberExpr>(target->IgnoreParenCasts())) {
+                AddDataFlow(targetFieldRef->getMemberDecl(), source);
+              }
+            }
+            void AddDataFlow(const ValueDecl* target, const Expr* source) {
+              auto* var = &db->ToVar(parent->GetUsr(target));
+              var->data_flow_into.push_back(GetDataFlow(source));
+            }
+            void AddDataFlow(const FunctionDecl* target, const Expr* source) {
+              auto* func = &db->ToFunc(parent->GetUsr(target));
```
MaskRay: Since we don't discourage uses of non-const references and `ToFunc` returns a non-const reference, we don't need to convert `func` to a pointer.

src/indexer.cc
```diff
@@ -830,6 +831,125 @@ class IndexDataConsumer : public index::IndexDataConsumer {
...
+                  write.role = Role::Call;
+                  return DataFlow{parent->GetUsr(data_flow_callee), write};
+                }
+              }
+              
+              return DataFlow{0,write};
+            }
+
+
+            IndexDataConsumer* parent;
```
MaskRay: `IndexDataConsumer *consumer;` may be better

src/messages/ccls_dataFlowInto.cc
```diff
@@ -0,0 +1,97 @@
...
+      const std::vector<DataFlow>* data_flow_sources = nullptr;
+      if (dataFlow.second == Kind::Var) {
+        auto& var = db->Var(dataFlow.first);
+        data_flow_sources = &var.data_flow_into;
+      } else if (dataFlow.second == Kind::Func) {
+        auto& func = db->Func(dataFlow.first);
+        data_flow_sources = &func.data_flow_into_return;
+      }
+      if (data_flow_sources != nullptr) {
+        if (data_flow_sources->size() == 0) {
```
MaskRay: `data_flow_sources->empty() `

src/query.cc
```diff
@@ -239,6 +243,8 @@ void DB::ApplyIndexUpdate(IndexUpdate *u) {
                  Use &use, int delta) {
     use.file_id =
         use.file_id == -1 ? u->file_id : lid2fid.find(use.file_id)->second;
+    if (usr == 0) return;
```
MaskRay: When can `usr == 0` happen?

src/messages/ccls_dataFlowInto.cc
```diff
@@ -0,0 +1,97 @@
...
+
+#include "message_handler.hh"
+#include "pipeline.hh"
+#include "query.hh"
+
+#include <unordered_set>
+
+namespace ccls {
+namespace {
+struct Param : TextDocumentPositionParam {
```
MaskRay: There is no need to define a new struct.

That was the purpose of commit 123f504d50e451bf84f71e109f230f220928fa4f. Before it, each method required a different parameter type.

src/query.cc
LeszekSwirski: When the data flow is from a non-trivial expression, which is considered a 'leaf' in the recursion.

src/messages/ccls_dataFlowInto.cc
LeszekSwirski: Then we'd have to move its REFLECT_STRUCT into message_handler.hh, otherwise the reflection below won't work.

Yanpas:
Vscode extension is ready for this feature:
https://github.com/Yanpas/vscode-ccls/commits/dataflow

PR #147, opened 2018-12-04
Spin IncludeComplete's destructor until scanning completes

LeszekSwirski:
Fixes a segfault when trying to access the group matcher in the scanning thread after IncludeComplete has been destroyed


MaskRay:
Thanks! Merged with a small clean-up. This file may have gotten the least revamp, but since it'll go away when wo get to clang 8.. i'm ok with the imperfect fix.

PR #159, opened 2018-12-13
Rendezvous after receiving "exit" notification

MaskRay:
@lhmouse

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -77,6 +77,7 @@ void StandaloneInitialize(MessageHandler &, const std::string &root);
 
 namespace pipeline {
 
+std::atomic<bool> quit;
 std::atomic<int64_t> loaded_ts = ATOMIC_VAR_INIT(0),
                      pending_index_requests = ATOMIC_VAR_INIT(0);
 int64_t tick = 0;
@@ -90,6 +91,10 @@ struct Index_Request {
   RequestId id;
   int64_t ts = tick++;
 };
+
+std::mutex thread_mtx;
+std::condition_variable no_pending_threads;
+int pending_threads;
 
 MultiQueueWaiter *main_waiter;
 MultiQueueWaiter *indexer_waiter;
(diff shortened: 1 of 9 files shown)


src/pipeline.cc
```diff
@@ -361,8 +366,31 @@ bool Indexer_Parse(SemaManager *completion, WorkingFiles *wfiles,
   return true;
 }
 
+void Quit(SemaManager &manager) {
+  quit = true;
+  manager.Quit();
+
+  { std::lock_guard lock(index_request->mutex_); }
+  indexer_waiter->cv.notify_all();
```
hotpxl: It’s better to protect the notify with the mutex. If the thread is woken up for any reason, the signal would be lost

src/sema_manager.cc
```diff
@@ -432,6 +436,8 @@ void *CompletionMain(void *manager_) {
   set_thread_name("comp");
   while (true) {
     std::unique_ptr<SemaManager::CompTask> task = manager->comp_tasks.Dequeue();
+    if (pipeline::quit)
```
hotpxl: A full memory barrier is probably not necessary here?

MaskRay: `ThreadedQueue::{Dequeue,PushBack}` use a mutex internally, the access of `pipeline::quit` is guaranteed to be synchronized.

lhmouse: Yes a seq_cst barrier is an overkill. Consider the following:

```c++
if(pipeline::quit.load(std::memory_order_relaxed)
  ...
````
(or `memory_order_acquire` if it wasn't protected by a mutex).

src/pipeline.cc
lhmouse: The code is correct. It is a bit obscure, however.

src/sema_manager.cc
MaskRay: It seems everything is protected by a mutex. I can change every load of `quit` to use `std::memory_order_relaxed`

Issue #160, opened 2018-12-13
Not completing include directories set in .ccls


oblitum:
I'm using coc.nvim+ccls, it provides most headers for completion, but doesn't provide headers I set in `.ccls` flags, for example, I'm trying this:

```
%clang
-Wall
-Wextra
%c -std=gnu11
%cpp -pedantic
%cpp -std=c++17
%cpp -stdlib=libc++
%cpp -I/opt/src/GSL/include
%cpp -I/opt/src/range-v3/include
```

When I go about including `#include <rang|>`, I only get boost range headers from my system, even after `#include <range/|>`, simple no completion for `range/v3`. But code completion and diagnostics works for range-v3 symbols, just inclusion that fails.

MaskRay:
Close due to inactivity and I cannot reproduce.

oblitum:
Sorry, I didn't have time to test that suggestion yet (I just grab it from AUR). But I'd report it back.

oblitum:
Does it work for you with coc.nvim and like my testcase explains? Because just by using coc.nvim I can't make it work at all.

MaskRay:
I checked it with releases.llvm.org clang 7 not the AUR package.

oblitum:
I was referring to ccls from AUR, not clang. I mean, I'm just installing ccls by doing `aurman -S ccls`, not compiling it myself, for checking the snippet you pointed.

oblitum:
@MaskRay 

Same `.ccls`:

```
%clang
-Wall
-Wextra
%c -std=gnu11
%cpp -pedantic
%cpp -std=c++17
%cpp -stdlib=libc++
%cpp -I/opt/src/GSL/include
%cpp -I/opt/src/range-v3/include
```

I've put the following:

```cpp
    std::fstream f("/home/francisco/ccls-includes");
    for (auto &E : HeaderOpts.UserEntries) {
      std::string path =
          NormalizePath(ResolveIfRelative(entry.directory, E.Path));
      f << "user entry: " << E.Path << '\n';
      f << "normalized: " << path << '\n';
      f.flush();
```

`ccls-includes`:
```
user entry: /usr/local/include
normalized: /usr/local/include
user entry: lib/clang/7.0.1/include
normalized: /home/francisco/Projects/samples/c++/lib/clang/7.0.1/include
user entry: /include
normalized: /include
user entry: /usr/include
normalized: /usr/include
```

Notice how relative `lib/clang/7.0.1/include` became a bad normalized input.

[...]

PR #171, opened 2018-12-21
Extend .ccls

### LINKED ISSUE #115: false diagnostics in include files due to never-firing #ifdefs
Makefile:
```
c.o: c.c; gcc -o $@ -c $<
```

c.c:
```
#define PLEASE
#include "i.h"
```
i.h:
```
#ifndef PLEASE
#error say PLEASE
#endif
```
Run `bear make` to get a `compile_commands.json`, open `i.h` in an editor configured to work with lsp & ccls, see how the `#error say PLEASE` lights up (even though PLEASE is clearly defined in the only compilation that `i.h` participates in).

MaskRay:
Fix #115 @madscientist

* Add `%h` for C header files (the suffix `.h` is considered a C header, not a C++ header)
* Add `%hpp` for C++ header files
* If `.ccls` exists, it provides full command line for files not specified `compile_commands.json` (before, `compile_commands.json` was ignored)
* If the first line of .ccls is `%compile_commands.json`, it appends flags to `compile_commands.json` "arguments", instead of overriding.
  Files not specified by `compile_commands.json` will not be added to folder.entries, but their command line can be inferred from other files.

### DIFF
--- a/src/include_complete.cc
+++ b/src/include_complete.cc
@@ -61,39 +61,26 @@ size_t TrimCommonPathPrefix(const std::string &result,
 }
 
 // Returns true iff angle brackets should be used.
-bool TrimPath(Project *project, std::string &path) {
+int TrimPath(Project *project, std::string &path) {
   size_t pos = 0;
-  bool angle = false;
-  for (auto &[root, folder] : project->root2folder) {
-    size_t pos1 = 0;
-    for (auto &search : folder.angle_search_list)
-      pos1 = std::max(pos1, TrimCommonPathPrefix(path, search));
-    if (pos1 > pos) {
-      pos = pos1;
-      angle = true;
-    }
-
-    pos1 = TrimCommonPathPrefix(path, root);
-    for (auto &search : folder.quote_search_list)
-      pos1 = std::max(pos1, TrimCommonPathPrefix(path, search));
-    if (pos1 > pos) {
-      pos = pos1;
-      angle = false;
-    }
-  }
+  int kind = 0;
+  for (auto &[root, folder] : project->root2folder)
+    for (auto &[search, kind1] : folder.search_dir2kind)
+      if (int t = TrimCommonPathPrefix(path, search); t > pos)
+        pos = t, kind = kind1;
   path = path.substr(pos);
-  return angle;
+  return kind;
 }
 
 CompletionItem BuildCompletionItem(const std::string &path,
-                                   bool use_angle_brackets) {
+                                   int kind) {
   CompletionItem item;
   item.label = ElideLongPath(path);
   item.detail = path; // the include path, used in de-duplicating
   item.textEdit.newText = path;
   item.insertTextFormat = InsertTextFormat::PlainText;
-  item.use_angle_brackets_ = use_angle_brackets;
   item.kind = CompletionItemKind::File;
+  item.quote_kind_ = kind;
   item.priority_ = 0;
   return item;
 }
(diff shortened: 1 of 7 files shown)


madscientist:
Wow thanks for the quick response!  I'm testing now but something is not quite right.  At first I was getting an error in *Messages* (maybe while ccls was rebuilding the index?) but now I get no errors but the header doesn't appear to be included.  I do see the `--include` flag mentioned in the log though.  I'll do some debugging provide more details.

madscientist:
OK more details: it appears to be related to the combination of compile_commands.json and .ccls.  If I remove compile_commands.json and change my `.ccls` to have all the flags:
```
%c
-I/tmp/cclstest/inc
%h
--include=global.h
```
then it works!  If I put back compile_commands.json, even with the above .ccls, it doesn't work again.

BTW, I think a more formal definition of the syntax of `.ccls` in the wiki would be helpful; while the above works, if I put the option on the same line as `%h`
```
%h --include=global.h
```
then it doesn't work... is that intended?  Using `%c -I/tmp/cclstest/inc` on one line does seem to work.

MaskRay:
I have fixed `%h`. It required some more work.

`%c` `%h` only applies the flag (only one, read the source) on the same line.

```
%c -I/tmp/cclstest/inc
%h --include=global.h
```

madscientist:
Awesome, I'll give it a try.

I was confused about the precise syntax of the `.ccls` file; the https://github.com/MaskRay/ccls/wiki/Getting-started wiki page shows this syntax:

```
%clang
%c -std=gnu11
%cpp -std=gnu++14
%objective-c %objective-cpp -DGNUSTEP
%c %cpp -DFOO
-pthread

# Includes
-I/work/ccls/third_party
-I/work/ccls/another_third_party
# -I space_is_not_allowed
```
which implies that flags on a line by themselves are valid...?  Maybe I misunderstood it.

MaskRay:
With your comment https://github.com/MaskRay/ccls/issues/115#issuecomment-449455357 in mind, I've changed the semantic a bit.

For you `/tmp/cclstest` example to work, you may also use:

```
% cat .ccls
%compile_commands.json
%h --include=inc/global.h
```

See the updated description of the commit. The source is still the best way to understand what's going on :) `project.cc`

MaskRay:
@oblitum @bstaletic

oblitum:
This didn't fix inclusion of headers at `%cpp -I/opt/src/range-v3/include` for me :-/ (#160).

oblitum:
Can you give an example how `%h` and `%hpp` is supposed to be used? Is it for include directories or files?

oblitum:
> Can you give an example how `%h` and `%hpp` is supposed to be used? Is it for include directories or files?

Nevermind, I've noticed from discussion it's for files.

madscientist:
@MaskRay OK tested in my real code... it works great!!  Now all my headers have full type information.  Love it!

@oblitum  Right, for files.  It's for extra flags to be given to the compiler when ccls "compiles" header files alone, rather than as part of a source file.  I don't expect it will be the most commonly-used option but it's critical in some situations.

MaskRay:
@oblitum For clang < 8, `include_complete.cc` (inherited from cquery) is used. I've commented at https://github.com/MaskRay/ccls/issues/160#issuecomment-449589865 The latest `project` branch should address #160

oblitum:
@MaskRay I've checked, it's working now.

PR #176, opened 2018-12-25
Delay requests if the document has not not indexed

MaskRay:
This fixes a plethora of "not indexed" errors when the document has not been indexed.

* Message handler throws NotIndexed
* The message is put into backlog and tagged with backlog_path
* path2backlog keeps backlog associated with a document and a merged index clears the backlog
* backlog[0] is forced to run if it becomes overdue

@yshui @hotpxl


yshui:
LGTM

PR #177, opened 2018-12-25
Add missing unordered_map include to working_files.hh

t6:
Building 0.20181225 on FreeBSD fails with:

src/working_files.hh:89:8: error: no template named 'unordered_map' in namespace 'std'
  std::unordered_map<std::string, std::unique_ptr<WorkingFile>> files;
  ~~~~~^
src/working_files.cc:248:8: error: no member named 'unordered_map' in namespace 'std'
  std::unordered_map<uint64_t, int> hash_to_unique;
  ~~~~~^


MaskRay:
Thanks! It seems I tagged the pre-release too early :) Since I would make some other rearrangement, I'll amend that faulty commit.

Sorry for the inconvenience to you.

t6:
Thank you.  However it is a bad idea and produces more work for packagers if you forcefully retag things.  The tarball generated by GitHub changes checksums if you do, which I have to record in the FreeBSD port. It would have been better/simpler/more transparent to just create a new tag instead.

MaskRay:
Sorry I'll create a new tag (like 0.20181111.1) next time.

MaskRay:
Sorry I'll create a new tag (like 0.20181111.1) next time.

t6:
Thanks! :+1:

MaskRay:
And thank you for creating the port! 😍

PR #178, opened 2018-12-26
Correct pending_threads counter

CXuesong:
There are `ThreadLeave` calls without corresponding `ThreadEnter` calls in

* `PreambleMain`
* `CompletionMain`
* `DiagnosticMain`

This will cause `pending_threads` go negative, and eventually cause ccls unable to exit after receiving `exit` notification.
Correct exit criteria in `LaunchStdin`. `method` is empty after `move`.

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -423,6 +423,7 @@ void ThreadLeave() {
   std::lock_guard lock(thread_mtx);
   if (!--pending_threads)
     no_pending_threads.notify_one();
+  assert(pending_threads >= 0);
 }
 
 void Init() {
@@ -526,13 +527,13 @@ void LaunchStdin() {
       ReflectMember(reader, "method", method);
       if (method.empty())
         continue;
+      const auto should_exit = method == "exit";
       // g_config is not available before "initialize". Use 0 in that case.
       on_request->PushBack(
           {id, std::move(method), std::move(message), std::move(document),
            chrono::steady_clock::now() +
                chrono::milliseconds(g_config ? g_config->request.timeout : 0)});
-
-      if (method == "exit")
+      if (should_exit)
         break;
     }
   }).detach();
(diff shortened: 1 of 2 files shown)


MaskRay:
It may unclear but `SpawnThread` calls `ThreadEnter`. The three functions should not call `ThreadEnter`.

MaskRay:
I think the correct fix is like:

```patch
diff --git a/src/pipeline.cc b/src/pipeline.cc
index 78ecde80..818ae321 100644
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -519,24 +519,25 @@ void LaunchStdin() {
       JsonReader reader{document.get()};
       if (!reader.m->HasMember("jsonrpc") ||
           std::string((*reader.m)["jsonrpc"].GetString()) != "2.0")
-        return;
+        break;
       RequestId id;
       std::string method;
       ReflectMember(reader, "id", id);
       ReflectMember(reader, "method", method);
       if (method.empty())
         continue;
+      bool should_exit = method == "exit";
       // g_config is not available before "initialize". Use 0 in that case.
       on_request->PushBack(
           {id, std::move(method), std::move(message), std::move(document),
            chrono::steady_clock::now() +
                chrono::milliseconds(g_config ? g_config->request.timeout : 0)});
 
-      if (method == "exit")
+      if (should_exit)
         break;
     }
+    ThreadLeave();
   }).detach();
-  ThreadLeave();
 }
 
 void LaunchStdout() {
@@ -553,8 +554,8 @@ void LaunchStdout() {
       if (stdout_waiter->Wait(quit, for_stdout))
         break;
     }
+    ThreadLeave();
   }).detach();
-  ThreadLeave();
 }
 
 void MainLoop() {
```

CXuesong:
Well that explains why the usage of `ThreadLeave` appears to be fishy in the original commit. Thanks!

PR #214, opened 2019-01-19
log to stderr to forward to lsp client

Yanpas:
This allows to forward logs to LSP client:
image

### DIFF
--- a/src/log.cc
+++ b/src/log.cc
@@ -67,11 +67,14 @@ Message::Message(Verbosity verbosity, const char *file, int line)
 }
 
 Message::~Message() {
-  if (!file)
-    return;
   std::lock_guard<std::mutex> lock(mtx);
   stream_ << '\n';
-  fputs(stream_.str().c_str(), file);
+  if (file) {
+    fputs(stream_.str().c_str(), file);
+    fflush(file);
+  }
+  fputs(stream_.str().c_str(), stderr);
+  fflush(stderr); // stderr is redirected to LSP client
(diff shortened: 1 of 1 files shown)


MaskRay:
I'm not sure what various language clients/servers do regarding logs. On many POSIX systems (not AIX), you can do `-log-file=/dev/stderr`

Riatre:
Unfortunately there is one operating system which is not POSIX and has no /dev/stderr equivalent despite still supports magical filenames such as "NUL" "CON" "PRN" from 80s: Windows.

Yanpas:
Works on Windows and Linux. Overhead of this change is low since ccls doesn't log to frequent.

Yanpas:
And it's absolutely not obvious that if you want to see log in your language client you have to write this launch argument. I've discovered it randomly

MaskRay:
@Riatre If there isn't a pseudo device for `/dev/stderr` on Windows, we may use `stderr` as a ccls-local notation and use `-log-file=stderr` to do what `-log-file=/dev/stderr` does on POSIX systems.

We can probably make `-log-file=stderr` the default. What do you think?

Yanpas:
Me? I agree

Riatre:
SGTM

That said, I believe "add -log-file=stderr and make it default" is effectively same as @Yanpas 's solution, and his solution is simpler.

MaskRay:
Merged #219 

Added `fflush` as @Yanpas suggested.

I prefer it as the log goes to one place.

PR #216, opened 2019-01-19
Allow Clang version and architecture to be overridden.

madscientist:
CMake only allows you to override cached variables on the command line.  Convert the Clang version and architecture to cache variables so they can be set via the cmake command line (if you want to use Clang 7.0.1 or some different architecture such as `x86_64-linux-gnu-ubuntu-18.04`, without having to edit files).

### DIFF
--- a/cmake/DownloadAndExtractClang.cmake
+++ b/cmake/DownloadAndExtractClang.cmake
@@ -6,27 +6,40 @@
 # Downloads 7-Zip to extract Clang if it isn't available in the PATH
 function(download_and_extract_clang CLANG_DOWNLOAD_LOCATION)
 
-set(CLANG_VERSION 7.0.0)
-set(CLANG_ARCHIVE_EXT .tar.xz)
+set(CLANG_VERSION 7.0.0
+    CACHE STRING "Version of Clang to download")
 
 if(${CMAKE_SYSTEM_NAME} STREQUAL Linux)
 
   # Default to Ubuntu 16.04
-  set(CLANG_ARCHIVE_NAME
-      clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-16.04)
+  set(_default_CLANG_ARCH x86_64-linux-gnu-ubuntu-16.04)
 
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL Darwin)
 
-  set(CLANG_ARCHIVE_NAME clang+llvm-${CLANG_VERSION}-x86_64-apple-darwin)
+  set(_default_CLANG_ARCH x86_64-apple-darwin)
 
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
 
-  set(CLANG_ARCHIVE_NAME LLVM-${CLANG_VERSION}-win64)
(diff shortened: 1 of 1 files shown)


MaskRay:
`clang_archive_hashes/` contains SHA256 hashes of some prebuilt archives. `-DCLANG_VERSION=7.0.1` may not be available currently.

madscientist:
Hrm, that's a good point.  I'm downloading it separately, actually, so I can avoid the download time when I rebuild ccls, so it doesn't compare the sha and works for my situation.  But you're right for a real solution something needs to be done about the hashes.  I'll look into it.

MaskRay:
If the user doesn't want to download prebuilt archives automatically (via cmake magic), they can download and unpack it manually and run:

`cmake -GNinja -H. -BRelease -DSYSTEM_CLANG=ON -DCMAKE_PREFIX_PATH=clang+llvm-7.0.0-x86_64-linux-gnu-ubuntu-14.04`

`-DSYSTEM_CLANG=` originates from some waf (wscript) logic and the name is less relevant now... Perhaps it should be renamed

madscientist:
`SYSTEM_CLANG` is a little misleading i agree.  Maybe the control of the download should be named `DOWNLOAD_CLANG` or something like that instead.

Also, `SYSTEM_CLANG` has other effects besides downloading: it will enable/disable RPATH setting which I think is orthogonal to downloading: probably this IS related to whether you're using system Clang or not, but just because you're not downloading doesn't mean you're using system Clang.

Finally, `CMAKE_PREFIX_PATH` is problematic because it doesn't play well (in this case) with `CMAKE_FIND_ROOT_PATH`; the latter is used when you have a cross-compilation environment or a separate sysroot (which we do) but it impacts `CMAKE_PREFIX_PATH` (paths there are appended to `CMAKE_FIND_ROOT_PATH` which is not what we want.  Probably sometimes you want it but not in this case.

However I did discover the cmake files support `CLANG_ROOT` which is an alternate location to search for clang libraries and that works OK with `CMAKE_FIND_ROOT_PATH`, so I can use that instead of `CMAKE_PREFIX_PATH`.

MaskRay:
@DaanDeMeyer knows all the internals...

DaanDeMeyer:
I mostly added the clang downloading to keep parity with the existing build system in cquery. Without that constraint I wouldn't implement it and require users to provide their own clang installation (or statically link libclang and provide binaries).

I'd consider removing support for downloading clang from CMake and instead adding instructions on how users can get an up-to-date version of clang installed (which shouldn't be all too difficult for most distro's). The CMake build would be dramatically simplified as a result as well.

MaskRay:
@DaanDeMeyer Thank you for offering your insights! I also appreciate simplicity and just created https://github.com/MaskRay/ccls/pull/221

MaskRay:
Merged #221 (which deletes the auto-download mechanism and `SYSTEM_CLANG`)

The burden for users who don't have convenient system clang+llvm libraries shouldn't be too much: i've updated ccls/wiki to mention that they need to download and unpack the releases.llvm.org prebuilt archives by themselves.

Still, thanks for improving the build system! (inspired #221)

PR #227, opened 2019-01-24
Simplify CMake by using LLVM Cmake config files

DaanDeMeyer:
Simplifies the CMake build even further by using the LLVM CMake config files instead of implementing the logic ourselves.

I couldn't use the LLVM CMake config files in cquery because they aren't bundled with the LLVM Windows archive just like the LLVM libraries but since you use the LLVM libraries it's already impossible to use ccls with the Windows archive which means we might as well assume the LLVM CMake config files are available.

I verified ccls builds with these changes but I don't use it myself so you might want to check if everything still works as expected before merging this.


MaskRay:
Thanks for your help!

`-I` for clang/llvm headers isn't specified after the simplification.

```zsh
LLVM=$(realpath ~/llvm)
cmake -H. -Bbuild -G Ninja -DCMAKE_PREFIX_PATH="$LLVM/Release;$LLVM/Release/tools/clang;$LLVM;$LLVM/tools/clang"
```

`ninja -C build`:
```
../src/serializer.hh:20:10: fatal error: 'llvm/Support/Compiler.h' file not found
#include <llvm/Support/Compiler.h>
```

DaanDeMeyer:
Did you `ninja -C build install` after building LLVM? I think the LLVM CMake config files assume you've installed LLVM and don't work directly from the build directory. That might be another downside of this approach. 

The config files themselves should be located in `<prefix>/lib/cmake` after installing.

DaanDeMeyer:
@MaskRay Any interest in merging this? If not I'll close the PR.

MaskRay:
Thank you for your help! I've created #313 based on #227. It removes (unrelated) formatting part from this PR and adjusts the logic a bit.

```cmake
find_package(LLVM REQUIRED)
find_package(Clang REQUIRED)   # this is also needed
```

I figured out the include directory issue. It requires:
```cmake
target_include_directories(ccls PRIVATE "src;${LLVM_INCLUDE_DIRS};${CLANG_INCLUDE_DIRS}")
```

madscientist:
Technically you don't need the quotes or semicolons (they kind of cancel each other out in your implementation).  This would work (although either way does work):

```target_include_directories(ccls PRIVATE  src ${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS})```

**Comment by DaanDeMeyer** [comment]
I'm surprised `target_include_directories` is needed at all. The include directories should get added automatically by `target_link_libraries`.

**Comment by MaskRay** [comment]
@madscientist Thanks. Changed to

```cmake
target_include_directories(ccls PRIVATE src)
target_include_directories(ccls SYSTEM PRIVATE
  third_party ${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS})
```

> I'm surprised target_include_directories is needed at all. The include directories should get added automatically by target_link_libraries.

@DaanDeMeyer I don't understand how `target_include_directories` `target_link_libraries` work, but `${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS}` are required for this case.

PR #236, opened 2019-01-29
stdin stdout logging, debug level logging

Yanpas:
image

### DIFF
--- a/src/log.cc
+++ b/src/log.cc
@@ -29,6 +29,13 @@ static std::mutex mtx;
 FILE *file;
 Verbosity verbosity;
 
+bool SetVerbosity(int value) {
+  verbosity = static_cast<Verbosity>(value);
+  if (value > Verbosity_DEBUG || value < Verbosity_FATAL)
+    return false;
+  return true;
+}
+
 Message::Message(Verbosity verbosity, const char *file, int line)
     : verbosity_(verbosity) {
   using namespace llvm;
@@ -60,6 +67,7 @@ Message::Message(Verbosity verbosity, const char *file, int line)
     case Verbosity_ERROR: stream_ << 'E'; break;
     case Verbosity_WARNING: stream_ << 'W'; break;
     case Verbosity_INFO: stream_ << 'I'; break;
+    case Verbosity_DEBUG: stream_ << 'D'; break;
     default: stream_ << "V(" << int(verbosity_) << ')';
   }
   // clang-format on
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -35,6 +35,7 @@ limitations under the License.
 
 #include <chrono>
 #include <mutex>
+#include <charconv>
 #include <shared_mutex>
 #include <thread>
 #ifndef _WIN32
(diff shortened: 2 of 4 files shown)


src/pipeline.cc
```diff
@@ -413,6 +414,43 @@ void Quit(SemaManager &manager) {
...
+  std::string buf;
+  while (true) {
+    int c = getchar();
+    if (c == EOF) {
+      *eof = true;
+      return;
+    }
+    if (c == '\n') {
+      if (buf.empty()) {
+        if (headers->content_length == 0)
```
MaskRay: What client did you see that may have such problems?

Yanpas: I was writing tool for system tests and ccls was silent

src/pipeline.cc
```diff
@@ -413,6 +414,42 @@ void Quit(SemaManager &manager) {
...
+          LOG_S(WARNING) << "Content-Length was not defined in header";
+        return;
+      }
+      if (!buf.compare(0, kContentLength.size(), kContentLength)) {
+        std::from_chars_result result = std::from_chars(
```
MaskRay: Unclear why the plain old atoi should be replaced by the much more complicated from_chars...

src/pipeline.cc
```diff
@@ -483,52 +520,53 @@ void LaunchStdin() {
...
+        LOG_S(WARNING) << "Json parse error";
+        continue;
+      }
 
       JsonReader reader{document.get()};
       if (!reader.m->HasMember("jsonrpc") ||
-          std::string((*reader.m)["jsonrpc"].GetString()) != "2.0")
-        break;
+          std::string((*reader.m)["jsonrpc"].GetString()) != "2.0") {
+        LOG_S(WARNING) << "Bad jsonrpc member";
```
MaskRay: The error checking for jsonrpc may not be that useful.

src/log.cc
```diff
@@ -60,6 +67,7 @@ Message::Message(Verbosity verbosity, const char *file, int line)
     case Verbosity_ERROR: stream_ << 'E'; break;
     case Verbosity_WARNING: stream_ << 'W'; break;
     case Verbosity_INFO: stream_ << 'I'; break;
+    case Verbosity_DEBUG: stream_ << 'D'; break;
```
MaskRay: `LOG_V(1)` already does what `LOG_V(DEBUG)` will do..

MaskRay:
How about https://github.com/MaskRay/ccls/tree/dev

src/pipeline.cc
Yanpas: I preserve previous behaviour AND log reason if the request is skipped

src/log.cc
Yanpas: DEBUG is more readable than 1. It's unclear what 1 means.

Yanpas:
What's wrong with my PR? The main purpose of it is tracing server. Log levels I guess should be documented somewhere.

MaskRay:
There are already `LOG_V(1)` and `LOG_V(2)` (2 isn't used currently) and thus `LOG_S(DEBUG)` may seem less useful. I understand that the word "DEBUG" is mnemonic but the verbosity level cannot be easily controlled with it. I use `-v=1` for my `ccls.debug` script which adds a bit more information, and `-v=2` when I really need more.

The error checking used in stdin/stdout handling is similar. I agree that they are not useless but the cases are rare and I do not intend to complicate the code a bit just to be a better checker of a random language client. The information can be easily obtained with other mechanisms, e.g. strace/truss/sysdig/dtrace. Bailing out when some error conditions happen isn't too bad, either. The server isn't designed to be fool-proof and I think it can crash to simplify processing in many places. If the client does something weird, just take them as undefined behavior.

Yanpas:
Maybe the code has become a little more complicated but much easier to read, since log messages are self-documenting, which results in simpler code for newcomers. If I still didn't persuade you - then I'm ok with changes in dev branch, just add some info about log levels from -3 to 2 (which is verbose and which isn't) please.

MaskRay:
The description of `-v=` has been updated in https://github.com/MaskRay/ccls/tree/dev

PR #301, opened 2019-03-04
Add clang.excludeGlobs option

xzz53:
This option allows to disable compilations options matching any of specified glob patterns. '*', '?', '[' and ']' metacharacters are supported. One use case is cross-compiling linux kernel: `ccls  -index=. '-init={"clang" : {"excludeGlobs" : ["-m*", "-W*", "-f*"]}}'` correctly indexes kernel sources using unmodified `compile_commands.json` obtained from `bear make`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ cquery has system include path detection (through running the compiler driver) w
 # >>> [Getting started](../../wiki/Home) (CLICK HERE) <<<
 
 * [Build](../../wiki/Build)
-* [Client feature table](../../wiki/Client-feature-table)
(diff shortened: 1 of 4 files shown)


MaskRay:
Rather than add a new option, we can make `clang.excludeArgs` to accept glob or regex. Have you measured the time spent on regex matching if these globc are changed to regex? (libstdc++ regex can be very slow (https://github.com/MaskRay/vscode-ccls/pull/25))

MaskRay:
https://github.com/MaskRay/ccls/pull/301/files

Issue #242, opened 2019-02-02
Receiving "code -32600" / message "not indexed"


madscientist:
I've started getting these errors back from ccls in my lsp-log buffer (Emacs 26.1, lsp-mode 20190126.1601, ccls-20181225.853, company-lsp-20190126.1750) for my large project.  I haven't been able to reproduce this in smaller sample codebases, but I've used a stripped-down Emacs init.el and can still reproduce it.  I've definitely waited for the entire codebase to index.  In fact this same problem happens if I enable `initialBlacklist (".")` to avoid indexing the entire project at once.

Nothing is completing, no symbols are resolved, etc.  I've enabled logging with `-v=1` and there are no errors or strange messages, the compile lines for my source are shown in the log and all look correct to me, and there is no problem finding the Clang resource directory for headers like stddef.h etc.

I'm not sure what this error code means or what to do next to debug further.

I get things like this in `*lsp-log*` in Emacs:

```<<<< ccls:7496
{
  "jsonrpc": "2.0",
  "id": 26,
  "error": {
    "code": -32600,
    "message": "not indexed"
  }
}
<<<< ccls:7496
{
  "jsonrpc": "2.0",
  "id": 27,
  "error": {
    "code": -32600,
    "message": "not indexed"
  }
}
<<<< ccls:7496
{
  "jsonrpc": "2.0",
  "id": 28,
  "result": {
    "signatures": [],
    "activeSignature": 0,
    "activeParameter": 0
  }
}
```

MaskRay:
A `"not indexed"` message is emitted here at https://github.com/maskray/ccls/blob/master/src/message_handler.cc#L279

The backlog mechanism is documented in the description of https://github.com/MaskRay/ccls/commit/0b80a10ca178eaee0bd572158728a478e4bc5ae7

When the main thread handles a message, it leaves `backlog` false. The first `"not indexed"` (`backlog==false`) throws an exception rather than issue an `ErrorMessage`.

Can you raise the initialization option `request.timeout` (defaults to 5000 milliseconds)?

MaskRay:
The `"not indexed"` `ResponseError` may be emitted either

* The requests takes more than 5 seconds because the file hasn't been indexed yet. This may happen (but rarely) for files with a pile of `#include`s.
[...]

PR #304, opened 2019-03-05
always use complete path as index lookup

s4c5:
When using ccls from emacs LSP, I would constantly get "not indexed" errors.
The source code being indexed is on a GPFS filesystem that has a "friendly" pathname, and the actual (more convoluted) gpfs pathname.
ccls would index the file under the convoluted pathname, then try to retrieve it using the friendly pathname, resulting in constant "not indexed" errors. By forcing ccls to use the full pathname with this patch, it will always uses the same key to reference the source code index, and always gets the correct result.
ccls was completely borken for me and with this patch it is working flawlessly. I would guess that this multiple-pathname problem would also be an issue for anyone with symlinks in their projects, this patch should fix those cases as well.

### DIFF
--- a/src/utils.cc
+++ b/src/utils.cc
@@ -113,7 +113,17 @@ std::string LowerPathIfInsensitive(const std::string &path) {
     c = tolower(c);
   return ret;
 #else
-  return path;
+  // use this opportunity to get full path
+  char *rpbuf = NULL;
+  std::string fullpath;
+  rpbuf = realpath(path.c_str(), NULL);
+  if (rpbuf != NULL) {
+    fullpath = rpbuf;
+    free(rpbuf);
+  } else {
+    fullpath = path;
+  }
+  return fullpath;
 #endif
 }
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ cquery has system include path detection (through running the compiler driver) w
 # >>> [Getting started](../../wiki/Home) (CLICK HERE) <<<
 
 * [Build](../../wiki/Build)
-* [Client feature table](../../wiki/Client-feature-table)
 * [FAQ](../../wiki/FAQ)
 
 ccls can index itself (~180MiB RSS when idle, noted on 2018-09-01), FreeBSD, glibc, Linux, LLVM (~1800MiB RSS), musl (~60MiB RSS), ... with decent memory footprint. See [wiki/compile_commands.json](../../wiki/compile_commands.json) for examples.
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -473,7 +473,8 @@ public:
       info.usr = HashUsr(USR);
       if (auto *ND = dyn_cast<NamedDecl>(D)) {
         info.short_name = ND->getNameAsString();
-        info.qualified = ND->getQualifiedNameAsString();
+        llvm::raw_string_ostream OS(info.qualified);
+        ND->printQualifiedName(OS, GetDefaultPolicy());
         SimplifyAnonymous(info.qualified);
       }
     }


MaskRay:
Adding `realpath()` logic to `LowerPathIfInsensitive` is not a correct fix. Can you give me the stacktrace when `LowerPathIfInsensitive` receives a non-absolute path argument?

s4c5:
```
#0  ccls::MessageHandler::FindOrFail (this=0x7fffffff12a0, path=..., reply=..., out_file_id=0x0) at /home/s4c5/emacs_support/ccls/src/message_handler.cc:279
#1  0x000000000058d3a2 in ccls::MessageHandler::textDocument_codeAction (this=0x7fffffff12a0, param=..., reply=...)
    at /home/s4c5/emacs_support/ccls/src/messages/textDocument_code.cc:38
#2  0x0000000000479b95 in void ccls::MessageHandler::Bind<ccls::CodeActionParam>(char const*, void (ccls::MessageHandler::*)(ccls::CodeActionParam&, ccls::ReplyOnce&))::{lambda(ccls::JsonReader&, ccls::ReplyOnce&)#1}::operator()(ccls::JsonReader&, ccls::ReplyOnce&) const (this=0x7fffffff12a0, reader=..., reply=...)                                         
    at /home/s4c5/emacs_support/ccls/src/message_handler.cc:165
#3  0x0000000000484fb3 in std::_Function_handler<void (ccls::JsonReader&, ccls::ReplyOnce&), void ccls::MessageHandler::Bind<ccls::CodeActionParam>(char const*, void (ccls::MessageHandler::*)(ccls::CodeActionParam&, ccls::ReplyOnce&))::{lambda(ccls::JsonReader&, ccls::ReplyOnce&)#1}>::_M_invoke(std::_Any_data const&, ccls::JsonReader&, ccls::ReplyOnce&) (
    __functor=..., __args#0=..., __args#1=...) at /physical/gpfs/rloc-grpusers/rloc_grp_users01/modules/linux/mods_linux/gcc-8.1.0/include/c++/8.1.0/bits/std_function.h:297
#4  0x000000000047b47b in std::function<void (ccls::JsonReader&, ccls::ReplyOnce&)>::operator()(ccls::JsonReader&, ccls::ReplyOnce&) const (this=0x1d35c38, __args#0=...,
    __args#1=...) at /physical/gpfs/rloc-grpusers/rloc_grp_users01/modules/linux/mods_linux/gcc-8.1.0/include/c++/8.1.0/bits/std_function.h:687
#5  0x000000000046edc3 in ccls::MessageHandler::Run (this=0x7fffffff12a0, msg=...) at /home/s4c5/emacs_support/ccls/src/message_handler.cc:223
#6  0x0000000000498ad8 in ccls::pipeline::MainLoop () at /home/s4c5/emacs_support/ccls/src/pipeline.cc:630
#7  0x000000000041dfa5 in main (argc=2, argv=0x7fffffff2088) at /home/s4c5/emacs_support/ccls/src/main.cc:158
(gdb) 
```
This is a stacktrace to a FindOrFail when it is looking for the "friendly" non-absolute pathname. The lookup fails because the "full" filename was used to index the file.
You can see the two different filenames referenced in the log:
```
14:45:57 indexer1         pipeline.cc:375 I store index for /physical/gpfs/rloc-grpusers/rloc_grp_users01/home_files/grp/s4c5/dev/baseproj/mainproj/mainproj/srcfile.cpp (delta: 0)                                                                                                                                                                            
14:46:00 ccls         message_handler.cc:279 I *NARF* file not found /home/s4c5/dev/baseproj/mainproj/mainproj/srcfile.cpp
```

I can do additional logging or more breakpoints as needed

madscientist:
Hm... this might be the same problem I'm seeing with issue #242 (still happening as of this past weekend although I totally understand why you closed it given that I've not had anything useful to say).  I just realized that my Git repository path contains a symlink to a different location.  Maybe I'll try without that and see if I still get the same behavior.

src/utils.cc
```diff
@@ -113,7 +113,17 @@ std::string LowerPathIfInsensitive(const std::string &path) {
     c = tolower(c);
   return ret;
 #else
-  return path;
+  // use this opportunity to get full path
+  char *rpbuf = NULL;
+  std::string fullpath;
+  rpbuf = realpath(path.c_str(), NULL);
```
myrgy: I would say that since project is based on c++17 it worth to use `std::filesystem`

https://en.cppreference.com/w/cpp/filesystem/absolute

MaskRay:
* The project root directory may be a symlink. I can add some code to always use the project root directory as passed from the language client as-is. If `/tmp/symlink/` resolves to `/tmp/real/` but the former is the root directory, always convert `/tmp/real/a.cc` to `/tmp/symlink/a.cc`. Will that fix your issue?
* Files may be symlinks. This is a tricky case as clang internals follow symlinks and the original filenames may be lost.

s4c5:
> Hm... this might be the same problem I'm seeing with issue #242 
You can see evidence of the problem if you look in the .ccls-cache directory in your project.
When I was seeing the pathname-aliasing problem, I had multiple "equivalent" paths showing up in the .ccls-cache. One would have the indexed data, the other would be empty.

s4c5:
> I can add some code to always use the project root directory as passed from the language client as-is.  Will that fix your issue?

It'd be worth a try. I haven't tracked down where the two different path versions are coming from. I suppose its possible that the client itself references the file by different pathnames at various times when communicating with ccls, in which case ccls would still need to normalize the two paths.

MaskRay:
See my updated comment in https://github.com/MaskRay/ccls/pull/304#issuecomment-470148610

MaskRay:
Superseded by https://github.com/MaskRay/ccls/pull/314

PR #306, opened 2019-03-07
use local rapidjson if available and drop git submodule

FirstLoveLife:
1. bump cmake to 3.11 for [FetchContent](https://cmake.org/cmake/help/v3.11/module/FetchContent.html)
2. if rapidjson is available already(e.g., `sudo pacman -S rapidjson` or build from source), then use it
3. else, fetch it with `FetchContent`

Pros:
1. https://www.reddit.com/r/cpp/comments/9284h5/dependency_management_with_cmake_and_git/e33xxh2/
2. avoiding git clone rapidjson if possible(as explained above)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,13 @@
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.11)
 project(ccls LANGUAGES CXX)
+
+### Colored output
+if(NOT WIN32)
+  string(ASCII 27 Esc)
+  set(Green       "${Esc}[32m")
+  set(BoldBlue    "${Esc}[1;34m")
+endif()
+
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/)
 include(DefaultCMakeBuildType)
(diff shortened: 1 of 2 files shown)


MaskRay:
The cmake on my machine is 3.10 :( 

Does the following cmake code do something similar to `git submodule update --init`?

```cmake
 FetchContent_Declare(
    rapidjson
    GIT_REPOSITORY https://github.com/Tencent/rapidjson
    GIT_TAG master
    )
```

FirstLoveLife:
> The cmake on my machine is 3.10 :(

It's a sad story.

> Does the following cmake code do something similar to git submodule update --init?

Yes.

MaskRay:
For fresh clone, I believe users can do `git clone https://github.com/MaskRay/ccls --recursive --depth=1 -j 2` ...

CMakeLists.txt
```diff
@@ -92,12 +114,48 @@ endif()
 target_compile_definitions(ccls PRIVATE
                            DEFAULT_RESOURCE_DIRECTORY=R"\(${Clang_RESOURCE_DIR}\)")
 
+
+### use local installed RapidJSON or FETCH from git via cmake
+find_package(RapidJSON QUIET)
```
MaskRay: This locates the system rapidjson (say /usr/include/rapidjson) if exists?

CMakeLists.txt
```diff
@@ -1,6 +1,28 @@
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.11)
 project(ccls LANGUAGES CXX)
 
+### Colored output
+if(NOT WIN32)
+  string(ASCII 27 Esc)
+  set(ColourReset "${Esc}[m")
+  set(ColourBold  "${Esc}[1m")
+  set(Red         "${Esc}[31m")
```
MaskRay: This doesn't need to be a full listing. Listing what are used should be good enough

CMakeLists.txt
```diff
@@ -92,12 +114,48 @@ endif()
...
                            DEFAULT_RESOURCE_DIRECTORY=R"\(${Clang_RESOURCE_DIR}\)")
 
+
+### use local installed RapidJSON or FETCH from git via cmake
+find_package(RapidJSON QUIET)
+if(RapidJSON_FOUND)
+  message("${BoldBlue}\nRapidJSON has been installed by your system before cloning ccls\n\n${ColourReset}")
+  include_directories(${RapidJSON_INCLUDE_DIRS})
+else()
+  message("${Green}====> Downloading RapidJSON ---- A fast JSON parser/generator for C++ with both SAX/DOM style API http://rapidjson.org/
```
MaskRay: I am not sure if FetchContent_Declare is better than `git submodule update --init` or `git clone --recursive`: it looks more opaque and the term "downloading" would actually worry me: in general I don't like the build system downloads something without my attention.

CMakeLists.txt
FirstLoveLife: Yes: https://github.com/Tencent/rapidjson/blob/master/RapidJSONConfig.cmake.in#L13

CMakeLists.txt
FirstLoveLife: > it looks more opaque and the term "downloading" would actually worry me: in general I don't like the build system downloads something without my attention.

fair enough. ccls uses only one git submodule, so `git submodule update update --init` is very close to `FetchContent`. So the only pros left is to use local rapidjson when possible. Generally speaking, FetchContent and git submodule are both fine for me. Feel free to close this pr if you don't like FetchContent. 

As for the term "downloading", I cannot think of better alternatives haha.

CMakeLists.txt
FirstLoveLife: removed

MaskRay:
https://cmake.org/files/

 | v3.10/ | 2018-03-16 09:38 | - |  
  | v3.11/ | 2018-06-14 13:50 | - |  
  | v3.12/ | 2018-11-02 13:51 | - |  
  | v3.13/ | 2019-02-01 13:20 | - |  
  | v3.14/ | 2019-03-08 11:09 | - |  
  | vCVS/ | 2019-03-09 01:20 | - |  

What do you think of https://github.com/MaskRay/ccls/pull/308 ?

FirstLoveLife:
#308 looks great, thanks!

PR #307, opened 2019-03-07
Hide all function declarations in documentSymbol.

CXuesong:
In current implementation of `textDocument/documentSymbol`, for function symbols of same {USR, Kind} in the same file, the returned function symbol will randomly be either one of the declarations, or the defintion. Consider a file containing both declaration(s) and implementation of a function. Because there is no way to distinguish between function declaration and its defintion with `ccls::Kind::Func` enum, {USR, Kind} pair collides in `sym2ds` map in `MessageHandler::textDocument_documentSymbol`.

When {USR, Kind} pair collides, the first seen symbol in `file->symbol2refcnt` dictionary wins and gets returned. This causes instability in `documentSymbol` result. For now, removing all the function declarations while keeping all the function defintions can make documentSymbol result determinstic (so is the "document outline" content in VSCode).

Further improvements, such as also showing function declarations in documentSymbol, IMO, depends on some sort of refactor of `textDocument_documentSymbol`. Perhaps we may allow `FuncDecl`s of same function to appear multiple times in the same file, and choose the correct `FuncDecl`/`Func` as the child of a Type considering the LsRange? It's only a rought thought.

### DIFF
--- a/src/indexer.hh
+++ b/src/indexer.hh
@@ -45,7 +45,7 @@ using Usr = uint64_t;
 
 // The order matters. In FindSymbolsAtLocation, we want Var/Func ordered in
 // front of others.
-enum class Kind : uint8_t { Invalid, File, Type, Func, Var };
+enum class Kind : uint8_t { Invalid, File, Type, Func, FuncDecl, Var };
 REFLECT_UNDERLYING_B(Kind);
 
 enum class Role : uint16_t {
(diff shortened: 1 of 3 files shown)


MaskRay:
```cpp
- enum class Kind : uint8_t { Invalid, File, Type, Func, Var };
+ enum class Kind : uint8_t { Invalid, File, Type, Func, FuncDecl, Var };
```

This would cause `-Wswitch` issues everywhere. I don't think adding a subcategory for this enum class is good.

```cpp
struct SymbolRef {
  Range range; // [0,8)
  Usr usr; // [8,16)
  Kind kind; // [16,17)
  Role role; // [18,20)

// ExtendRef::extent; // [20,24)
```

We can probably make use of the byte 17 (padding after `kind`).

MaskRay:
Superseded by https://github.com/MaskRay/ccls/pull/309

PR #326, opened 2019-03-19
Fix for "Change containers of Query*::Def fields from std::vector to ccls::Vec" not building with MVS

SysRay:
I couldn't build the project with Visual Studio 2017 and the latest changes from "Change containers of Query*::Def fields from std::vector to ccls::Vec".  The compiler complaint about the missing constructors and operators.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -20,11 +20,21 @@ Config *g_config;
 
 void DoPathMapping(std::string &arg) {
   for (const std::string &mapping : g_config->clang.pathMappings) {
-    auto colon = mapping.find(':');
+
+    auto colon = mapping.find('>');
     if (colon != std::string::npos) {
       auto p = arg.find(mapping.substr(0, colon));
       if (p != std::string::npos)
         arg.replace(p, colon, mapping.substr(colon + 1));
+    } 
+    else {
+      // Deprecated: Use only for older settings
+      auto colon = mapping.find(':');
+      if (colon != std::string::npos) {
+        auto p = arg.find(mapping.substr(0, colon));
+        if (p != std::string::npos)
+          arg.replace(p, colon, mapping.substr(colon + 1));
+      }
     }
   }
 }
--- a/src/indexer.hh
+++ b/src/indexer.hh
@@ -143,6 +143,9 @@ void Reflect(BinaryWriter &visitor, SymbolRef &value);
 void Reflect(BinaryWriter &visitor, Use &value);
 void Reflect(BinaryWriter &visitor, DeclRef &value);
 
+template <typename T>
+using VectorAdapter = std::vector<T, std::allocator<T>>;
+
 template <typename D> struct NameMixin {
   std::string_view Name(bool qualified) const {
     auto self = static_cast<const D *>(this);
@@ -156,7 +159,8 @@ template <typename D> struct NameMixin {
   }
 };
 
-struct FuncDef : NameMixin<FuncDef> {
+template <template <typename T> class V>
+struct FuncDef : NameMixin<FuncDef<V>> {
   // General metadata.
   const char *detailed_name = "";
   const char *hover = "";
@@ -164,11 +168,11 @@ struct FuncDef : NameMixin<FuncDef> {
   Maybe<DeclRef> spell;
 
   // Method this method overrides.
-  std::vector<Usr> bases;
+  V<Usr> bases;
   // Local variables or parameters.
-  std::vector<Usr> vars;
+  V<Usr> vars;
   // Functions that this function calls.
-  std::vector<SymbolRef> callees;
+  V<SymbolRef> callees;
 
   int file_id = -1; // not serialized
   int16_t qual_name_offset = 0;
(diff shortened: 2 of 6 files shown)


MaskRay:
> The compiler complaint about the missing constructors and operators.

Can you paste the build log?

SysRay:
Of course:
[build.txt](https://github.com/MaskRay/ccls/files/2984530/build.txt)

...
C:\Program Files\LLVM\include\clang/Basic/Module.h(128): note: see reference to class template instantiation 'llvm::SmallSetVector<const clang::FileEntry *,2>' being compiled
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\xmemory0(881): error C2280: 'std::pair<ccls::Usr,ccls::TypeDef<ccls::Vec>>::pair(const std::pair<ccls::Usr,ccls::TypeDef<ccls::Vec>> &)': attempting to reference a deleted function
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\utility(194): note: see declaration of 'std::pair<ccls::Usr,ccls::TypeDef<ccls::Vec>>::pair'
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\utility(194): note: 'std::pair<ccls::Usr,ccls::TypeDef<ccls::Vec>>::pair(const std::pair<ccls::Usr,ccls::TypeDef<ccls::Vec>> &)': function was implicitly deleted because a data member invokes a deleted or inaccessible function 'ccls::TypeDef<ccls::Vec>::TypeDef(const ccls::TypeDef<ccls::Vec> &)'
  f:\source\ccls\src\indexer.hh(226): note: 'ccls::TypeDef<ccls::Vec>::TypeDef(const ccls::TypeDef<ccls::Vec> &)': function was implicitly deleted because a data member invokes a deleted or inaccessible function 'ccls::Vec<ccls::Usr>::Vec(const ccls::Vec<ccls::Usr> &)'
  f:\source\ccls\src\utils.hh(154): note: 'ccls::Vec<ccls::Usr>::Vec(const ccls::Vec<ccls::Usr> &)': function was implicitly deleted because a data member invokes a deleted or inaccessible function 'std::unique_ptr<T [],std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)'
...

MaskRay:
I think this is a bug in MSVC but we may have to work around it.

Do you know which line causes the issue?

If you delete those `emplace_back` calls from query.cc, does it compile?
```cpp
      r.funcs_removed.emplace_back(func.usr, Convert(func.def));
...
      r.funcs_def_update.emplace_back(it.first, Convert(func.def));

```

MaskRay:
https://github.com/MaskRay/ccls/issues/327#issuecomment-474773200

SysRay:
Same error with:
     r.funcs_removed.emplace_back(func.usr, Convert(func.def));
...
      r.funcs_def_update.emplace_back(it.first, Convert(func.def));

MaskRay:
Thank you for reporting this. For that GCC bug I I'll go with:

```cpp
template <typename T> struct Vec {
  std::unique_ptr<T[]> a;
  int s = 0;
#if !(__clang__ || __GNUC__ > 7 || __GNUC__ == 7 && __GNUC_MINOR__ >= 4)
  // Work around a bug in GCC<7.4 that optional<IndexUpdate> would not be
  // construtible.
  Vec() = default;
  Vec(const Vec &o) : a(std::make_unique<T[]>(o.s)), s(o.s) {
    std::copy(o.a.get(), o.a.get() + o.s, a.get());
  }
  Vec(Vec &&) = default;
  Vec &operator=(Vec &&) = default;
  Vec(std::unique_ptr<T[]> a, int s) : a(std::move(a)), s(s) {}
#endif
  const T *begin() const { return a.get(); }
  T *begin() { return a.get(); }
  const T *end() const { return a.get() + s; }
  T *end() { return a.get() + s; }
  int size() const { return s; }
  const T &operator[](size_t i) const { return a.get()[i]; }
  T &operator[](size_t i) { return a.get()[i]; }
};
```

I guess the problem is around `src/pipeline.cc:664`

```
    bool indexed = false;
    for (int i = 20; i--;) {
      std::optional<IndexUpdate> update = on_indexed->TryPopFront();
```

Can you comment out this for loop and see it `pipeline.cc` compiles without the change to `Vec`?

SysRay:
Still not working with the same error.

MaskRay:
Can you be more specific what (MSVC 2017) doesn't work? gcc<7.4 cannot instantiate `std::optional<IndexUpdate>` without the user-defined ctors.

SysRay:
Sorry, commenting out the for loop doesn't help.

The workaround is doing fine, therefore I closed the request.

MaskRay:
`#if !(__clang__ || __GNUC__ > 7 || __GNUC__ == 7 && __GNUC_MINOR__ >= 4)`

If the compiler is clang, or the gcc version is greater or equal to 7.4, I think it doesn't need to user-defined ctors. `__clang__` is only defined by clang.

I would also like to know what is the minimum version of MSVC that can compile `pipeline.cc` without these user-defined ctors in `utils.hh`, and filter out those versions of MSVC.

PR #331, opened 2019-03-21
Allow an absolute file path for compilationDatabaseDirectory

SysRay:
If the build directory with the compilationDatabase.json is at a completely different place than root, you had to point to it relatively to root.
Accepting an absoulute file path is much cleaner.

### DIFF
--- a/src/project.cc
+++ b/src/project.cc
@@ -46,6 +46,7 @@ limitations under the License.
 #include <limits.h>
 #include <unordered_set>
 #include <vector>
+#include <filesystem>
 
 using namespace clang;
 using namespace llvm;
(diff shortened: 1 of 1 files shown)


MaskRay:
`compilationDatabaseDirectory` can be an absolute path now. So this is patch is not necessary.

SysRay:
I used an absolute path for ccls.misc.compilationDatabaseDirectory completely.
Example: 
     root: C/asd/code
     compilationDatabaseDirectory: Z:/home/user/build

CDBDir = root;
 if (g_config->compilationDatabaseDirectory.size()) {
      sys::path::append(CDBDir, g_config->compilationDatabaseDirectory); 
...

Result was: CDBDir = C:/asd/code/Z:/home/user/build/"compile_commands.json"

Did I miss something?

PR #332, opened 2019-03-21
PathMappings don't work on windows with ':' as seperator

SysRay:
Example: "C:/asd/:F:/dsa/"
Now:       "C:/asd/>F:/dsa/"

Or maybe a different seperator?


MaskRay:
@Riatre What's your opinion here?

Riatre:
I believe changes like this is needed for `pathMapping` on Windows because ':' is drive separator on Windows. Using '>' is okay for Windows because Windows tried very hard to prevent users from naming a file / directory with `\/:*?"<>`, but other platforms accept '><' in filename. Maybe make this Windows only?

SysRay:
I used an object for the path mappings {"C:/path1/" : "F:/path2/" , ...}. But all the plugins for the editors have to be changed as well, then.
Maybe check if it is an object and if not then just use ':' as an separator? So, it stays compatible.

MaskRay:
Shall we just use `>` as the separator on all platforms? `clang.pathMappings` is known to few people and the compatibility isn't a big deal.

myrgy:
what if use >, but allow user to define other symbol if required?

MaskRay:
Changed to use `>` on all platforms. Thanks you for reporting this issue!

```cpp
void DoPathMapping(std::string &arg) {
  for (const std::string &mapping : g_config->clang.pathMappings) {
    auto colon = mapping.find('>');
    if (colon != std::string::npos) {
      auto p = arg.find(mapping.substr(0, colon));
      if (p != std::string::npos)
        arg.replace(p, colon, mapping.substr(colon + 1));
    }
  }
}
```

PR #340, opened 2019-03-23
 improvge and fix UB of the use of log

FirstLoveLife:
https://github.com/MaskRay/ccls/blob/master/src/main.cc#L90 
```c++
  ccls::log::verbosity = ccls::log::Verbosity(opt_verbose.getValue());
```
Given the current `enum Verbosity` range from `-3` to `-0`, `Log_V(1)` and `Log_V( 2)` are not reliable because `1` and `2` are out of range.
According to CWG 1766:

> A value of integral or enumeration type can be explicitly converted to an enumeration type. The value is unchanged if the original value is within the range of the enumeration values (10.2 [dcl.enum]). Otherwise, <del>the resulting value is unspecified (and might not be in that range)</del> behavior is undefined.

The solution may be:

- `Message(Verbosity verbosity, const char* file, int line);` ->   `Message(int verbosity, const char* file, int line);`

or

- 
```c++
enum Verbosity {
  Verbosity_FATAL = -3,
  Verbosity_ERROR = -2,
  Verbosity_WARNING = -1,
  Verbosity_INFO = 0,
  // more level...
};
```

IMO the latter is better option.

Besides, this pr also replace macros inside log.hh with short functions and replace `enum Verbosity` with `enum class Verbosity`


MaskRay:
To be honest I didn't know CWG 1766 before you mention it. (Thanks!)

However, is `LOG_V(2)` UB? 

Quoting `[dcl.enum]`

> 8. For an enumeration whose underlying type is fixed, the values of the enumeration are the values of the underlying type. Otherwise, the values of the enumeration are the values representable by a hypothetical integer type with minimal width M such that all enumerators can be represented. The width of the smallest bit-field large enough to hold all the values of the enumeration type is M. It is possible to define an enumeration that has values not defined by any of its enumerators. If the enumerator-list is empty, the values of the enumeration are as if the enumeration had a single enumerator with value 0.99

```cpp
enum Verbosity {
  Verbosity_FATAL = -3,  // the hypothetical integer type has at least 3 bits (-3 => 0b101)
  Verbosity_ERROR = -2,
  Verbosity_WARNING = -1,
  Verbosity_INFO = 0,
};
```

The minimum 3 bits allow us to use log levels til `LOG_V(3)`.

FirstLoveLife:
Oops, my bad. I misunderstood the definition of range here

Issue #334, opened 2019-03-21
ccls --version reports (apparently) only the LLVM version


madscientist:
### Observed behavior

Running `ccls --version` shows just the LLVM version info:
```
$ ccls --version
ccls version 7.0.1 (tags/RELEASE_701/final)
```

### Expected behavior

Showing the LLVM version is useful, but I'd really like to see information on the version of ccls (e.g., the output of `git describe --tag`).  It's not always the case that I have ready access to the ccls Git repository that I built from when I need to know the version of ccls I'm using.

madscientist:
I can create a PR for this over the weekend if that would be OK.

MaskRay:
Thank you! That will be very useful!

MaskRay:
Fixed by #342

PR #341, opened 2019-03-23
add --version support

FirstLoveLife:
https://github.com/MaskRay/ccls/issues/334


yssource:
why are you hard coding the version?

FirstLoveLife:
@yssource I agree assigning it to a constexpr variable or leave it for cmake(like version.cmake.in)  are better choices, but it is not used anywhere else, so maybe it ain't that bad here.

yssource:
@FirstLoveLife I suggest it may get the latest version from **cmake add_custom_command**, then pass it as constexpr variable.

add_custom_command may invoke some codes like the following.
`git show -s --format=%ci `git log -n1 --format=format:"%H"``

MaskRay:
Ideally some cmake code runs `git describe --tags` and generates a macro definition in a `.hh` file which is included by `main.cc`......

yssource:
There is no problem if the user compile ccls from the official git source.
But, it would be a problem if the user install it by different package installer. Such as pacman installer, yum installer etc, since there is no git commit.
So, I think it still have to get the version from a fixed value.

madscientist:
I am working on the cmake version, related to issue #334  that I filed yesterday.

If you don't want me to work on that anymore it's fine, let me know.

PR #342, opened 2019-03-23
Enhance the output of --version with Git describe results.

madscientist:
Implement a solution to issue #334.  This also allows `-DCCLS_VERSION` to be provided via the `cmake` command line, for upstream facilities that want to set the version during the build without having access to the Git repository.

Note that this only obtains a new version when you run cmake, so if you do a `git pull` it will no longer have the correct version unless you rerun `cmake`.  I'm trying to create an implementation which always checks but only rebuilds if something changes; unfortunately it's very tricky to do this in CMake (I could do it in 5 seconds in make :smile:)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,3 +209,20 @@ target_sources(ccls PRIVATE
   src/messages/textDocument_signatureHelp.cc
   src/messages/workspace.cc
 )
+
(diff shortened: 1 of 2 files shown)


MaskRay:
I can do that in 5 minutes in make but not in cmake☺

PR #350, opened 2019-03-29
Fix invalid handling of llvm MemoryBuffer in DiagnosticMain.

davidwelch158:
When 'Buf' is passed to clang in addRemappedFile then clang will destroy it when the compiler instance is destroyed.
So release ownership of the pointer immediately.
Otherwise if the call to Parse fails then we destroy both the compiler instance (indirectly destroying the MemoryBuffer) and destroy the memory buffer again via the reference to it from 'Buf' which isn't cleared yet.
This can cause a segmentation fault.

Unfortunately I don't have a test case for this as I've stopped being to reproduce the error return from 'Parse'.


MaskRay:
Good catch! What command line did you use to make `Parse` fail?

MaskRay:
Can you try #351 ?

davidwelch158:
Yes, #351 fixes the issue. Unfortunately I couldn't reproduce the error from Parse today (even though it was happened reproducibly yesterday) - sorry about. But if I make manually fail in the debugger then ccls works with that change and crashes without.

PR #371, opened 2019-04-06
Add WIN32 to the check for a C++ stdlib workaround

lanza:
This same bug occurs with clang-cl master (as of 4/6/2019). Add WIN32
to the list of checks to opt into the workaround behavior.


MaskRay:
Is it due to any of `-fms-extensions -fms-compatibility -fms-compatibility-version=19.11 -fdelayed-template-parsing` (enabled by default for clang-cl)? Can you figure out which `-f` option causes the failure?

Riatre:
(I think the problem is more likely caused by clang-cl using MSVC's STL implementation.)

MaskRay:
If it is caused by MSVC's STL. It'd be helpful to know which version has fixed the issue.

Beside, the macro `_WIN32` is probably more common than `WIN32`.

PR #376, opened 2019-04-12
[WIP] Preliminary CUDA support

rietmann-nv:
This PR adds support for CUDA files.

The way CMake compiles cuda files happens in a way that makes it hard for `ccls` to parse the command and extract the relevant flags to make `clang` work correctly. I've added a few changes to `ccls` to add explicit CUDA support in both the file detection and `.ccls` configuration to make basic things like code completion and jump-to-definition mostly work.

The nvcc command looks like this (notice two step command)
```
/usr/local/cuda/bin/nvcc -ccbin=gcc-6  -I../src -I../external/cutlass -I../external/cub -isystem=../external/googletest/googletest/include   -Xcompiler -fopenmp --expt-extended-lambda --std=c++11 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_70,code=compute_70 -g   -x cu -c /home/max/dev/cuml/ml-prims/test/add.cu -o test/CMakeFiles/mlcommon_test.dir/add.cu.o && /usr/local/cuda/bin/nvcc -ccbin=gcc-6  -I../src -I../external/cutlass -I../external/cub -isystem=../external/googletest/googletest/include   -Xcompiler -fopenmp --expt-extended-lambda --std=c++11 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_70,code=compute_70 -g   -x cu -M /home/max/dev/cuml/ml-prims/test/add.cu -MT test/CMakeFiles/mlcommon_test.dir/add.cu.o -o $DEP_FILE
```
A similar clang command should look something like this:
```
clang -I../src -I../external/cutlass -I../external/cub -isystem=../external/googletest/googletest/include --std=c++11 --cuda-gpu-arch=sm_70 --cuda-path=/usr/local/cuda-9.2/ -c add.cu
```

I wanted this for myself and my work, but I know there are others who would like this as well. I wanted to make sure to contribute back, given how much I've been using `ccls` in the last year. Most of the additions are pretty simple, but the changes I made at `project.cc:437` feel a bit hacky, and I wanted your feedback to streamline them into something better, if that is possible. Basically I implemented a whitelist which only allows (`-I, -isystem, -std=`) type flags. This basically works, but maybe you have a better idea.

Additionally it requires the user to make their `.ccls` look something like:
```
%compile_commands.json
%cu --cuda-gpu-arch=sm_70
%cu --cuda-path=/usr/local/cuda-9.2/
```
The `--cuda-gpu-arch=sm_70` and `--cuda-path=/usr/local/cuda-9.2/` could possibly be added automatically.

Anyway, thanks for `ccls` and I look forward to your feedback.

### DIFF
--- a/src/project.cc
+++ b/src/project.cc
@@ -51,6 +51,15 @@ using namespace clang;
 using namespace llvm;
 
 namespace ccls {
+
+template <typename Enumeration>
+auto as_integer(Enumeration const value)
+  -> typename std::underlying_type<Enumeration>::type
+{
+  return static_cast<typename std::underlying_type<Enumeration>::type>(value);
+}
+
+
 std::pair<LanguageId, bool> lookupExtension(std::string_view filename) {
   using namespace clang::driver;
   auto I = types::lookupTypeForExtension(
@@ -59,7 +68,9 @@ std::pair<LanguageId, bool> lookupExtension(std::string_view filename) {
                 I == types::TY_ObjCXXHeader;
   bool objc = types::isObjC(I);
   LanguageId ret;
-  if (types::isCXX(I))
+  if (I == types::TY_CUDA)
+    ret = LanguageId::CUDA;
+  else if (types::isCXX(I))
     ret = objc ? LanguageId::ObjCpp : LanguageId::Cpp;
   else if (objc)
     ret = LanguageId::ObjC;
@@ -123,6 +134,8 @@ struct ProjectProcessor {
             ok |= lang == LanguageId::ObjC;
           else if (A.consume_front("%objective-cpp "))
             ok |= lang == LanguageId::ObjCpp;
+          else if (A.consume_front("%cu "))
+            ok |= lang == LanguageId::CUDA;
           else
             break;
         }
@@ -132,6 +145,7 @@ struct ProjectProcessor {
         args.push_back(arg);
       }
     }
+    
     entry.args = args;
     GetSearchDirs(entry);
   }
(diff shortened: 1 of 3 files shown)


src/project.cc
```diff
@@ -123,6 +134,8 @@ struct ProjectProcessor {
             ok |= lang == LanguageId::ObjC;
           else if (A.consume_front("%objective-cpp "))
             ok |= lang == LanguageId::ObjCpp;
+          else if (A.consume_front("%cu "))
```
MaskRay: `%cu ` is not necessary. CUDA is C++ (`clang::driver::types::isCXX` used in `lookupExtension`)

MaskRay:
Thank you for contribution. I don't have nvidia gpgpu :( so probably not able to check include-specific stuff.

ccls runs clang Lex/Parse/Sema, it doesn't run CodeGen; there is no need to specify linked libraries.
clang has CUDA support but its command line options may be a bit different from nvcc.  https://llvm.org/docs/CompileCudaWithLLVM.html In ccls, I really don't want to encode command line option scrubber. Instead, users are expected to scrub the options themselves, through the initialization options `clang.excludeArgs` and/or `compilationDatabaseCommand`.

Hard coding `--cuda-gpu-arch=sm_70` is bad. The only purpose of this option is to define the `__CUDA_ARCH__`. In clang, it doesn't do anything else.

MaskRay:
The `%cu` directive is probably necessary because these CUDA driver options should not be used in standard C++ files. I added support for it in commit 796c75794e2b524aec3bc7af2ce01b806c651b74. Thank you for the preliminary work:)

bavalpey:
@MaskRay Are there any plans to support cuda?  I notice that this wasn't merged, and ccls doesn't seem to support cuda...

MaskRay:
@bavalpey CUDA support has been available for a long time. Building CUDA applications with Clang requires some wrestling with include and library paths (https://llvm.org/docs/CompileCudaWithLLVM.html) but no ccls specific code is needed.

PR #383, opened 2019-04-27
Find rapidjson headers in rapidjson <= 1.1.0.

bastianbeischer:
The CMake variable name was renamed in

https://github.com/Tencent/rapidjson/commit/b91c515afea9f0ba6a81fc670889549d77c83db3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,6 +152,10 @@ if(USE_SYSTEM_RAPIDJSON)
 endif()
 if(NOT RapidJSON_FOUND)
   set(RapidJSON_INCLUDE_DIRS third_party/rapidjson/include)
+elseif(NOT RapidJSON_INCLUDE_DIRS)
+  # The variable name was changed in
+  # https://github.com/Tencent/rapidjson/commit/b91c515afea9f0ba6a81fc670889549d77c83db3
+  set(RapidJSON_INCLUDE_DIRS ${RAPIDJSON_INCLUDE_DIRS})
 endif()
 target_include_directories(ccls SYSTEM PRIVATE ${RapidJSON_INCLUDE_DIRS})
 


MaskRay:
Oh so, no stable release since [1.1.0 (Aug 25, 2016)](https://github.com/Tencent/rapidjson/releases/tag/v1.1.0)? I worry that there may be source adaptations ccls will have to do. That is what I want to avoid. So `third_party/include/rapidjson` (`git submodule update --init` or `git clone --recursive`) is probably recommended before they have a new release.

Thank you for noticing this.

bastianbeischer:
Right. rapidjson wasn't released for almost three years. I agree that using the thirdparty/include/rapidjson is indeed the best thing to do. I would say that the USE_SYSTEM_RAPIDJSON mode for released rapidjson versions should be fixed regardless, but that's up to you of course :)

By the way, there is also another problem in case rapidjson headers are in /usr/include: In that case -isystem /usr/include is added to the compiler flags which causes more problems.

Overall, I guess the USE_SYSTEM_RAPIDJSON should default to OFF, then? And the wiki should be extended so that the users make sure to checkout the rapidjson submodule before building. Or you could have CMake do it...

MaskRay:
The wiki page recommended `git clone --depth=1 --recursive https://github.com/MaskRay/ccls`. The `--recursive` option is to clone rapidjson into `third_party/rapidjson`. If the user fails to do that, `git submodule update --init`. Thank you for the information. I just made another +1 in one of the release request issues in the rapidjson repository.

```cmake
if(USE_SYSTEM_RAPIDJSON)   ## it is a no-op if a system rapidjson is not found
  find_package(RapidJSON QUIET)
endif()
```

bastianbeischer:
Sorry for not actually checking the wiki :-( You are right of course.

So then the only problem which remains is this: If they do a release of rapidjson and it is deployed in /usr (which is the normal way e.g. in Arch Linux), then ccls will actually add the include path to target_include_directories since RapidJSON_INCLUDE_DIRS is then set. But then ccls will not build because it uses -isystem /usr/include. But that problem wouldn't be fixed by this PR, so it needs a separate fix...

MaskRay:
Non-existent `-isystem` is just ignored. System rapidjson installs to `/usr/include`, which is already in the default search paths.

bastianbeischer:
Sorry, should have been more precise (again). Installing rapidjson-git to /usr/include will bring the following ${RapidJSON_INCLUDE_DIRS}:

-isystem /usr/lib64/cmake/RapidJSON/../../../include

(which is /usr/include). This gives (for example):

```
usr/bin/c++   -I/home/beischer/repos/ccls/src -isystem /home/beischer/repos/ccls/third_party -isystem /usr/lib64/cmake/RapidJSON/../../../include  -O3 -DNDEBUG   -Wall -Wno-sign-compare -Wno-return-type -Wno-unused-result -pthread -std=c++17 -o CMakeFiles/ccls.dir/src/clang_tu.cc.o -c /home/beischer/repos/ccls/src/clang_tu.cc
In file included from /usr/include/c++/8.3.0/ext/string_conversions.h:41,
                 from /usr/include/c++/8.3.0/bits/basic_string.h:6400,
                 from /usr/include/c++/8.3.0/string:52,
                 from /usr/include/c++/8.3.0/stdexcept:39,
                 from /usr/include/c++/8.3.0/optional:38,
                 from /home/beischer/repos/ccls/src/utils.hh:18,
                 from /home/beischer/repos/ccls/src/position.hh:18,
                 from /home/beischer/repos/ccls/src/clang_tu.hh:18,
                 from /home/beischer/repos/ccls/src/clang_tu.cc:16:
/usr/include/c++/8.3.0/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
 #include_next <stdlib.h>
               ^~~~~~~~~~
compilation terminated.
```

PR #384, opened 2019-04-28
Fix nested .ccls

Riatre:
In short, nested .ccls is broken in current HEAD. There are 2 issues:

1. At https://github.com/MaskRay/ccls/blob/e496d4e5db47b602b14fe53bd556f11d0b025f52/src/project.cc#L499 file paths are compared with folder path using startswith, consider the following case:

There are 3 `.ccls`:
1. `/a/b/root/.ccls` (folder: `/a/b/root`)
2. `/a/b/root/1/.ccls` (folder: `/a/b/root/1`)
3. `/a/b/root/11/.ccls` (folder: `/a/b/root/11`)

For file `/a/b/root/11/a.cc`, it matches all of these, which one will be used depends on the iterating order of `std::unordered_map`, i.e. pretty much random.

2. At
https://github.com/MaskRay/ccls/blob/e496d4e5db47b602b14fe53bd556f11d0b025f52/src/project.cc#L276
paths returned by `sys::path::parent_path` does not have trailing slashes, however project root used at https://github.com/MaskRay/ccls/blob/e496d4e5db47b602b14fe53bd556f11d0b025f52/src/project.cc#L265 does have (see https://github.com/MaskRay/ccls/blob/e496d4e5db47b602b14fe53bd556f11d0b025f52/src/project.cc#L455), which leads to a dummy `folder.dot_ccls` entry being inserted.

The logic in `FindEntry` is quite complicated, I don't think I understand the original problem here, so feel free to suggest better fixes.

### DIFF
--- a/src/project.cc
+++ b/src/project.cc
@@ -491,33 +491,38 @@ Project::Entry Project::FindEntry(const std::string &path, bool can_redirect,
 
   bool exists = false;
   std::string dir;
-  const std::vector<const char *> *extra = nullptr;
+  const std::vector<const char *> *args_from_dotccls = nullptr;
   Project::Entry ret;
+  std::string best_dotccls_root;
+  ccls::Project::Folder *best_dotccls_folder = nullptr;
   for (auto &[root, folder] : root2folder)
     if (StringRef(path).startswith(root))
       for (auto &[dir1, args] : folder.dot_ccls)
-        if (StringRef(path).startswith(dir1)) {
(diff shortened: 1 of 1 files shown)


MaskRay:
Thanks for suggesting the fix! I've merged the first commit. For the second one, I agree the logic is intertwined. I created https://github.com/MaskRay/ccls/pull/386 to simplify it and fix the problem. Does that work for you?

Riatre:
Closed in favor of #386

PR #386, opened 2019-04-29
Refactor FindEntry and use best-fit .ccls

MaskRay:
Fixes an issue with hierarchical .ccls found by @Riatre in #384

### DIFF
--- a/src/project.cc
+++ b/src/project.cc
@@ -240,8 +240,6 @@ std::vector<const char *> GetFallback(const std::string path) {
   std::vector<const char *> argv{"clang"};
   if (sys::path::extension(path) == ".h")
     argv.push_back("-xobjective-c++-header");
-  for (const std::string &arg : g_config->clang.extraArgs)
-    argv.push_back(Intern(arg));
   argv.push_back(Intern(path));
   return argv;
 }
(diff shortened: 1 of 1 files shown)


src/project.cc
```diff
@@ -470,94 +470,99 @@ void Project::Load(const std::string &root) {
...
-  ret.is_inferred = true;
-  ret.filename = path;
-  if (!best) {
-    if (ret.root.empty())
-      ret.root = g_config->fallbackFolder;
-    ret.directory = ret.root;
-    ret.args = GetFallback(path);
+  bool append;
+  if (best_dot_ccls_args && !(append = AppendToCDB(*best_dot_ccls_args))) {
+    // Override the compdb entry.
```
Riatre: It is stated on [wiki](https://github.com/MaskRay/ccls/wiki/Project-Setup#compile_commandsjson-1) that 
> By default .ccls compiler flags are applied only to files not listed in compile_commands.json.

However this is going to override the compile db entry.

MaskRay: This rule may be confusing. I think it is probably clearer to let `.ccls` override `compile_commands.json` if the first option is not `%compile_commands.json`

Riatre:
Yeah. Depends on actual use case both new or old behavior may be wanted
(e.g. "use compile_commands.json primarily and .ccls as fallback" vs "use
special -D flag configurations which is meaningless for actual builds in
some directories") so I have no opinion on this, both are okay for my use
case, just not sure if it is intended or not.

Fangrui Song <notifications@github.com> 于2019年4月30日周二 下午7:38写道：

MaskRay:
I've reinstated the old behavior. The user can just delete `compile_commands.json` (likely a symlink) to make `.ccls` win. I end up with more refactoring as it turns out that lazy `clang.extraArgs` appending improves readability.

PR #393, opened 2019-05-09
Fixed hierachicalPath on windows

SysRay:
### Problem
Paths on windows start with ":". So, a normal path like **c:/cache** is changed to c@/cache, which is invalid.

### Solution
Skipping the first 2 characters (the drive name and colon, c:).


MaskRay:
> Paths on windows start with ":". So, a normal path like c:/cache is changed to c@/cache, which is invalid.

I don't understand. Which path component is invalid on Windows?

SysRay:
For example: `g_config->cache.directory` is `C:/code/cache/`

```
std::string ret =  g_config->cache.directory + (src[0] == '/' ? src.substr(1) : src);
#ifdef _WIN32	
    std::replace(ret.begin(), ret.end(), ':', '@');
#endif
    return ret;
  }
```

Since all colons are replaced by @ on windows,  ret starts now with `C@/code/cache/` instead of originally `C:/code/cache/`.  
`C@/code/cache/` is at a completely different location. 

Expectation:
The desired cache location is not altered on windows.

MaskRay:
Thank you for reporting the issue. I think a better fix is:

```cpp
  if (g_config->cache.hierarchicalPath) {
    std::string ret = src[0] == '/' ? src.substr(1) : src;
#ifdef _WIN32
    std::replace(ret.begin(), ret.end(), ':', '@');
#endif
    return g_config->cache.directory + ret;
  }
```

Bump 0.20190314 to 0.20190314.1 to add the change.

PR #409, opened 2019-06-13
Avoid pessimizing moves in threaded_queue.hh

torokati44:
Clang produces warnings during compilation:

ccls/src/threaded_queue.hh:150:27: warning: moving a local object in a return statement prevents copy elision [-Wpessimizing-move]
ccls/src/threaded_queue.hh:150:27: note: remove ‘std::move’ call

This patch fixes the issue.

### DIFF
--- a/src/threaded_queue.hh
+++ b/src/threaded_queue.hh
@@ -147,7 +147,7 @@ public:
       auto val = std::move(q->front());
       q->pop_front();
       --total_count_;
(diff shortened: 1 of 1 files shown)


MaskRay:
Thanks! This was written this way to work around a gcc<~4.9 bug. Now we've switched to -std=c++17 and required GCC>=7.2, the workaround should not be used.

PR #417, opened 2019-06-20
Don't add include paths as SYSTEM

dcermak:
With the introduction of gcc 9 my builds of ccls on openSUSE Tumbleweed started failing on ppc, arm and SystemZ with odd error messages like the following:
```
[   58s] [ 11%] Building CXX object CMakeFiles/ccls.dir/src/fuzzy_match.cc.o
[   58s] /usr/bin/c++   -I/home/abuild/rpmbuild/BUILD/ccls-0.20190314/src -isystem /home/abuild/rpmbuild/BUILD/ccls-0.20190314/third_party -isystem /usr/lib64/cmake/RapidJSON/../../../include  -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -DNDEBUG -O2 -g -DNDEBUG   -Wall -Wno-sign-compare -Wno-return-type -Wno-unused-result -fno-rtti -pthread -std=c++17 -o CMakeFiles/ccls.dir/src/fuzzy_match.cc.o -c /home/abuild/rpmbuild/BUILD/ccls-0.20190314/src/fuzzy_match.cc
[   58s] In file included from /usr/include/c++/9/ext/string_conversions.h:41,
[   58s]                  from /usr/include/c++/9/bits/basic_string.h:6493,
[   58s]                  from /usr/include/c++/9/string:55,
[   58s]                  from /home/abuild/rpmbuild/BUILD/ccls-0.20190314/src/fuzzy_match.hh:19,
[   58s]                  from /home/abuild/rpmbuild/BUILD/ccls-0.20190314/src/fuzzy_match.cc:16:
[   58s] /usr/include/c++/9/cstdlib:75:15: fatal error: stdlib.h: No such file or directory
[   58s]    75 | #include_next <stdlib.h>
[   58s]       |               ^~~~~~~~~~
[   58s] compilation terminated.
```

I've tracked this down to being caused by the usage of `-isystem` for the include paths instead of `-I`.

This pull request removes the flag `SYSTEM` from the `target_include_directories()` calls, which fixes the compilation problems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,8 +144,7 @@ set_property(SOURCE src/utils.cc APPEND PROPERTY COMPILE_DEFINITIONS
 ### Includes
 
 target_include_directories(ccls PRIVATE src)
-target_include_directories(ccls SYSTEM PRIVATE
-  third_party ${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS})
+target_include_directories(ccls SYSTEM PRIVATE third_party)
 
 if(USE_SYSTEM_RAPIDJSON)
   find_package(RapidJSON QUIET)
@@ -153,7 +152,33 @@ endif()
 if(NOT RapidJSON_FOUND)
   set(RapidJSON_INCLUDE_DIRS third_party/rapidjson/include)
 endif()
-target_include_directories(ccls SYSTEM PRIVATE ${RapidJSON_INCLUDE_DIRS})
+
+# only add the include directories for LLVM, clang & rapidjson if they aren't in
+# the default list of include directories (these are stored by cmake in
+# CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES)
+#
+# Adding the same include directory twice with SYSTEM results these being
+# included as -isystem, which causes issues with libstdc++ from gcc9 with their
+# usage of #include_next<>
+# see also: https://github.com/MaskRay/ccls/pull/417
+foreach(include_dir
+    ${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS} ${RapidJSON_INCLUDE_DIRS})
+
+  # rapidjson's include dir is a relative path => normalize it
+  get_filename_component(include_dir_realpath ${include_dir} REALPATH)
+  set(dir_already_included FALSE)
+
+  foreach(cxx_include_dir ${CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES})
+    if("${cxx_include_dir}" STREQUAL "${include_dir_realpath}")
+      set(dir_already_included TRUE)
+    endif()
+  endforeach()
+
+  if(NOT "${dir_already_included}")
+    target_include_directories(ccls SYSTEM PRIVATE ${include_dir})
+  endif()
+endforeach()
+
 
 ### Install
 


MaskRay:
Thank you reporting the issue.

Changing `-isystem` to `-I` may give us warnings from llvm and rapidjson.

> #include_next <stdlib.h>
>
> -isystem /usr/lib64/cmake/RapidJSON/../../../include

The libstdc++ search path `/usr/include/c++/9` should be placed before `/usr/include` in the include path stack. However, this line will reorder `/usr/include` before the libstdc++ search path.

I don't remember why https://github.com/MaskRay/ccls/pull/35 added `SYSTEM`. @scturtle can probably give some insights.

I think this is a generic issue that should affect older versions of GCC as well, not sure what has changed in GCC 9.

scturtle:
I usually use `SYSTEM` to hide warnings from system or third-party headers that may not be fault of mine.

dcermak:
> I think this is a generic issue that should affect older versions of GCC as well, not sure what has changed in GCC 9.

Not necessarily, this could be caused by a change in libstdc++/STL only. I've looked into this a little further and the issue is the usage of `#include_next <stdlib.h>` in `cstdlib`: `#include_next` tells the compiler to include `cstlib` but to omit everything from the header file search path that was already searched to get to this header. Combining this with the different behavior of `-isystem` vs `-I`, this can result in headers not being found anymore. A more complete explanation can be found here: https://stackoverflow.com/a/37218954.

I agree that my fix is relatively ugly, as it will result in warnings showing up from standard headers. Unfortunately I don't know another way beside having to reordering the `target_include_directories()` calls, if the C++ headers change.

MaskRay:
A better approach it to check if `/usr/lib64/cmake/RapidJSON/../../../include` is in the default list of search paths `{gcc,clang} -v -fsyntax-only -xc++ /dev/null`. Unfortunately I don't know how to do this in cmake:

* resolve `/usr/lib64/cmake/RapidJSON/../../../include`  to `/usr/include`
* check if `/usr/include` is in the list

dcermak:
> A better approach is to check if `/usr/lib64/cmake/RapidJSON/../../../include` is in the default list of search paths `{gcc,clang} -v -fsyntax-only -xc++ /dev/null`. Unfortunately I don't know how to do this in cmake:

This should be possible by checking the contents of [CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_IMPLICIT_INCLUDE_DIRECTORIES.html). 

I can revise this PR to include such a check if you'd prefer that. We could add such a check for the addition of the clang & llvm include directories too, as these tend to be in the default include path anyway (at least on Linux if you install clang via the package manager).

MaskRay:
> I can revise this PR to include such a check if you'd prefer that. We could add such a check for the addition of the clang & llvm include directories too, as these tend to be in the default include path anyway (at least on Linux if you install clang via the package manager).

Please do that if you can figure out it. Thanks!

dcermak:
> Please do that if you can figure out it. Thanks!

Done.

Could you test this on other platforms, especially on Windows and with llvm ending up outside of the standard search path?

Riatre:
Builds fine for me on Windows (with GCC 9.1.0 and Clang 8 shipped by MSYS2).

MaskRay:
Reworded and simplified to:

```cmake
target_include_directories(ccls PRIVATE src)

foreach(include_dir third_party
    ${LLVM_INCLUDE_DIRS} ${CLANG_INCLUDE_DIRS} ${RapidJSON_INCLUDE_DIRS})
  get_filename_component(include_dir_realpath ${include_dir} REALPATH)
  # Don't add as SYSTEM if they are in CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES.
  # It would reorder the system search paths and cause issues with libstdc++'s
  # use of #include_next. See https://github.com/MaskRay/ccls/pull/417
  if(NOT "${include_dir_realpath}" IN_LIST CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES)
    target_include_directories(ccls SYSTEM PRIVATE ${include_dir})
  endif()
endforeach()
```

dcermak:
Thanks for merging this, the openSUSE builds of ccls are green now again!

Issue #418, opened 2019-06-20
ccls does not normalize/canonicalize workspace paths


eklitzke:
I have two indexed projects, `~/code/ccls` and `~/code/cruise`. In my ccls workspace cache I see four entries:
```
$ ls -l
total 9316
drwxrwx--- 2 eklitzke eklitzke   57344 Jun 20 11:38 @@home@eklitzke@code@ccls
drwxrwx--- 2 eklitzke eklitzke    4096 Jun 20 11:38 @home@eklitzke@code@ccls
drwxrwx--- 2 eklitzke eklitzke 7512064 Jun 19 12:54 @@home@eklitzke@code@cruise
drwxrwx--- 2 eklitzke eklitzke 1937408 Jun 19 14:22 @home@eklitzke@code@cruise
```

It appears this is because for some reason some of the paths of files are starting with `//` instead of just `/`, but these after normalization these are considered to be the same paths on Unix. The implementation of `NormalizePath()` in `platform_posix.cc` only removes dots, but not duplicated slashes:
```c++
std::string NormalizePath(const std::string &path) {
  llvm::SmallString<256> P(path);
  llvm::sys::path::remove_dots(P, true);
  return {P.data(), P.size()};
}
```

I'm not that familiar with the LLVM fs stuff, but I didn't see any methods to normalize or canonicalize paths. In C++17 there is `std::filesystem::canonical` that will handle this case, or else it would be trivial to write a small amount of code to handle this case.

MaskRay:
> It appears this is because for some reason some of the paths of files are starting with // instead of just /,

No, this is a deliberate choice to encode paths outside of workspace folders.

```cpp
// src/pipeline.cc
std::string GetCachePath(std::string src) {
  ...
  ///////// paths outside of workspace folders are encoded as @@path@to@file
  return g_config->cache.directory + '@' +
         EscapeFileName(g_config->fallbackFolder.substr(
             0, g_config->fallbackFolder.size() - 1)) +
         '/' + EscapeFileName(src);
}
```

Two projects may share some global include files. Because the included files may be included with different defined macros (think: `_FILE_OFFSET_BITS`), it is better for them not to share cache files, so these paths are prefixed with `@@path@to@project`.

> In C++17 there is std::filesystem::canonical that will handle this case, or else it would be trivial to write a small amount of code to handle this case.

[...]

PR #419, opened 2019-06-20
Use realpath(3) to canonicalize file paths on POSIX systems

eklitzke:
This is my proposed fix for #418. It's not fancy, but:
 * It's portable (`realpath(3)` is required by POSIX)
 * It's correct
 * It doesn't rely on bleeding edge C++17 `std::filesystem` stuff that's not yet ready in GCC/Clang
 * It works on paths longer than 256 characters 🤣

### DIFF
--- a/src/platform_posix.cc
+++ b/src/platform_posix.cc
@@ -39,9 +39,6 @@ limitations under the License.
 #include <malloc.h>
 #endif
 
-#include <llvm/ADT/SmallString.h>
-#include <llvm/Support/Path.h>
-
 #include <atomic>
 #include <condition_variable>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


eklitzke:
I guess I am missing something because this doesn't actually fix the issue. In `pipeline.cc` I see some logic in `GetCachePath()` that sometimes adds an additional `@` at the front of the path, and sometimes does not. With the change here, I still see ccls creating duplicate cache entries.

I don't understand what's supposed to be happening in this function, but I've confirmed that this change doesn't fix the underlying issue (although I content that the current logic still probably does not correctly handle paths like `/foo/////bar.txt`). Any ideas?

src/platform_posix.cc
```diff
@@ -52,10 +49,14 @@ namespace pipeline {
...
 }
 
+// Normalize the path, including eliminating redundant . and duplicated /
+// characters.
 std::string NormalizePath(const std::string &path) {
-  llvm::SmallString<256> P(path);
-  llvm::sys::path::remove_dots(P, true);
-  return {P.data(), P.size()};
+  char canonical[PATH_MAX + 1];
+  if (realpath(path.c_str(), canonical)) {
```
myrgy: Is it cross-platform? Seems like that function doesn't exist on windows.

MaskRay:
See https://github.com/MaskRay/ccls/issues/418#issuecomment-504260555

This is intentional :)

PR #423, opened 2019-06-23
Fix Getting started (CLICK HERE) link

duianto:
The text `(CLICK HERE)` was not clickable.


MaskRay:
This really... doesn't matter. I can change this next time when I have to touch README

PR #429, opened 2019-06-28
Caller Hiearchy shows lines of each reference in caller, not start li…

withgod84:
As like other IDEs (Eclipse, ..) , it's better to show line of each reference in caller hiearchy, 
Instead of caller's start line.
For example, if caller has a lots of lines, it's annoying to find which line calls current symbol in which context or condition.

### DIFF
--- a/src/messages/ccls_call.cc
+++ b/src/messages/ccls_call.cc
@@ -79,14 +79,14 @@ bool Expand(MessageHandler *m, Out_cclsCall *entry, bool callee,
   entry->numChildren = 0;
   if (!def)
     return false;
-  auto handle = [&](SymbolRef sym, int file_id, CallType call_type1) {
+  auto handle = [&](SymbolRef sym, Use use, CallType call_type1) {
     entry->numChildren++;
     if (levels > 0) {
       Out_cclsCall entry1;
       entry1.id = std::to_string(sym.usr);
       entry1.usr = sym.usr;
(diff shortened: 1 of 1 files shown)


withgod84:
Ready to merge in my opinion.

MaskRay:
* I remember someone suggested to use the function line instead of the call site for caller hierarchy. It may not be obvious from the UI which functions are the callers.
* src/messages/ccls_call.cc is used by both **caller** and **callee** hierarchies. This change will break callee hierarchy.
* `foo` may be called multiple times in `bar`. Creating duplicate nodes in the caller hierarchy is redundant.

withgod84:
> * I remember someone suggested to use the function line instead of the call site for caller hierarchy. It may not be obvious from the UI which functions are the callers.
> * src/messages/ccls_call.cc is used by both **caller** and **callee** hierarchies. This change will break callee hierarchy.
===> First of all, my change only shows function line only in case of caller hiearchy. 
===> In case of callee, it's better to show definition of callee, but it's not case in caller. 
===> IMHO, it's better to take different strategy between caller and callee. 

> * `foo` may be called multiple times in `bar`. Creating duplicate nodes in the caller hierarchy is redundant.
===> This is reasonable point, but this may harm efficiency of using caller hiearchy
===> For every entry in graph, user should find the location which call the function.

I think there is trade off between function line and reference line. 
If then, how about creating new parameter option to show reference line instead of function start line?

Issue #437, opened 2019-07-11
ccls not echoing ID from initialize request in response


thomasjm:
### Observed behavior

The LSP spec says that the `id` field of an `initialize` request should be included in the corresponding `ResponseMessage`. But when I send one, the `ResponseMessage` with the `InitializeResult` always seems to contain the string `"0"` for the ID.

### Expected behavior

The response to my `initialize` request should contain the request ID I sent.

### Steps to reproduce

1. Start the `ccls` language server
2. Send a valid `initialize` request with a nontrivial `id` and observe the response.

Maybe the ID is just not getting initialized and most LSP clients tolerate this?

### System information

* ccls version: I'm using the "Pi Day" release
* clang version: N/A
* OS: Linux
* Editor: N/A
* Language client (and version): custom client I'm writing

MaskRay:
> But when I send one, the ResponseMessage with the InitializeResult always seems to contain the string "0" for the ID.

You may have misread the request/response messages. If the `initialize` request specifies `"id":0`, the ResponseMessage will encode `"id":0`. It cannot be `string "0"`.

thomasjm:
@MaskRay the LSP spec allows request/response IDs to be either numbers or strings. So it had better be possible for it to be the string `"0"` or that would be another issue.

In fact, I think that's where the bug lies. I just switched my client to send numeric IDs instead of strings, and that makes it work.

So there is indeed a bug with handling string IDs.

Riatre:
I have read JSONRPC spec again, it indeed allows request ID to be either numbers or **arbitrary** strings. (I'd say this is pretty much fucked up though) 

ccls already tries to handle this, but it assumes string IDs are numeric, which explains those zeros you saw. If you really want arbitrary strings as ID, try the following patch:

```patch
diff --git a/src/lsp.cc b/src/lsp.cc
index 5c56f324..a41a809f 100644
--- a/src/lsp.cc
+++ b/src/lsp.cc
@@ -26,16 +26,16 @@ namespace ccls {
 void Reflect(JsonReader &vis, RequestId &v) {
   if (vis.m->IsInt64()) {
     v.type = RequestId::kInt;
-    v.value = int(vis.m->GetInt64());
+    v.value = std::to_string(int(vis.m->GetInt64()));
   } else if (vis.m->IsInt()) {
     v.type = RequestId::kInt;
-    v.value = vis.m->GetInt();
+    v.value = std::to_string(vis.m->GetInt());
[...]

PR #440, opened 2019-07-15
Apply patch from @Riatre for issue #437

thomasjm:
This is just applying @Riatre's patch for #437. It fixes the problem for me so merging it into `ccls` would be nice. Removing the "invalid" label from the issue would also be nice :P

### DIFF
--- a/src/lsp.hh
+++ b/src/lsp.hh
@@ -23,6 +23,7 @@ limitations under the License.
 
 #include <chrono>
 #include <iosfwd>
+#include <string>
 
 namespace ccls {
 struct RequestId {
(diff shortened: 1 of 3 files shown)


MaskRay:
Merged with a small tweak

```patch
-    w.String(id.value.c_str(), id.value.length());
-    w.String(id.value.c_str(), id.value.size());
```

> Removing the "invalid" label from the issue would also be nice :P

Will do :P

src/lsp.hh
```diff
@@ -31,7 +32,7 @@ struct RequestId {
   enum Type { kNone, kInt, kString };
```
myrgy: std::variant might be a good alternative

MaskRay: std::variant was used but removed in commit "Remove variant and clean up". If it is to be used in more places, it can be reconsidered.

PR #445, opened 2019-07-20
project: improve ComputeGuessScore heuristics

MaskRay:


src/project.cc
```diff
@@ -318,15 +318,40 @@ void LoadDirectoryListing(ProjectProcessor &proc, const std::string &root,
...
+    if (c == '/') {
+      score -= 9;
+      if (h)
+        m[h]++;
+      h = 0;
+    } else {
+      h = h * 33 + c;
+    }
+  h = 0;
+  for (uint8_t c : a)
```
betasheet: Should this iterate over b instead? :)

for (uint8_t c : b)

Issue #235, opened 2019-01-29
Save error on windows


Yanpas:
I do often save files which I edit, sometimes they are read by ccls and fails happen. As far as I remember WinAPI forbids reader and writer access to file. Maybe there are some winapi functions that fix it...
![image](https://user-images.githubusercontent.com/7345761/51941731-5da59280-2426-11e9-9413-ea6b3671ab8f.png)

MaskRay:
Interesting issue.

* client sends `textDocument/didSave`
* main thread (`messages/textDocument_did.cc`) enqueues an indexing task (`pipeline::Index`)
* indexer thread (`pipeline.cc`) reads the file to check if the file index (`IndexFile`) is stale. If it is, call into clangIndex to index the file (many (uncontrollable) files may be read via `#include`)

Do you know the order of `textDocument/didSave` and the actual save event of the file?

Yanpas:
I'm afraid I cannot. It happened when I hit save twice with 1 second interval

MaskRay:
You remind me of another thing: clang 8 have nanosecond precision for timestamps. I'll work on that.

MaskRay:
If vscode doesn't ensure `textDocument/didSave` is sent after the save action finishes, I think this can't be fixed on ccls' side.

Riatre:
Even if VSCode ensure `textDocument/didSave` is sent after the save action finished, this can still happen, nothing prevents me from saving the file again while ccls is parsing the file. On Windows, if the file is opened without some strange flags like `FILE_SHARE_READ | FILE_SHARE_WRITE`, the file will be effectively locked, all subsequent `CreateFile` calls will fail until the owner closes this file. I don't know if there are ways to configure how clang reads file though.

MaskRay:
If a failed `::CreateFileW(PathUTF16.begin(), Access, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, &SA, Disp, Flags, NULL);` (https://github.com/llvm-mirror/llvm/tree/master/lib/Support/Windows/Path.inc#L1097) call pops up a window, I cannot think of a way to fix it. It it doesn't do that and just returns a error number, there will be [a spurious `diag::err_pp_file_not_found` error](https://github.com/llvm-mirror/clang/tree/master/lib/Lex/PPDirectives.cpp#L1871):

```text
a.cc:1:10: fatal error: 'stdio.h' file not found
#include <stdio.h>
```

[...]

PR #453, opened 2019-07-30
Use volatile filesystem on Windows

Crunkle:
When Clang loads up the files required for indexing it memory maps any source files over a specific threshold. This is fine for Unix, but the strict file locking policies on Windows cause issues especially when an IDE saves the file and triggers a `textDocument/didSave`. If the file is re-saved before the previous indexing has complete, the save will fail and often produce various ambiguous errors.

I have been using this for a month and I believe clangd has something very similar implemented. No issues so far, even on very large source files, so I want to merge if possible.

### DIFF
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -1264,6 +1264,12 @@ Index(SemaManager *manager, WorkingFiles *wfiles, VFS *vfs,
   Clang->createFileManager();
 #endif
 
+#ifdef _WIN32
+  // Always set UserFilesAreVolatile true to avoid write locks on Windows
+  Clang->setSourceManager(new SourceManager(Clang->getDiagnostics(),
+                                            Clang->getFileManager(), true));
+#endif
+
   IndexParam param(*vfs, no_linkage);
   auto DataConsumer = std::make_shared<IndexDataConsumer>(param);
 


MaskRay:
Thanks for looking into the issue! It is probably https://github.com/MaskRay/ccls/issues/235

> When Clang loads up the files required for indexing it memory maps any source files over a specific threshold.

To expand this topic a bit, `clang::FrontendAction::BeginSourceFile` creates a SourceManager with UserFilesAreVolatile set to true if it doesn't exist:

```cpp
// lib/Frontend/FrontendAction.cpp#L673
  if (!CI.hasSourceManager())
    CI.createSourceManager(CI.getFileManager());

// lib/Frontend/CompilerInstance.cpp
void CompilerInstance::createSourceManager(FileManager &FileMgr) {
  SourceMgr = new SourceManager(getDiagnostics(), FileMgr); // third arg defaults to false
}
```

> This is fine for Unix, but the strict file locking policies on Windows cause issues especially when an IDE saves the file and triggers a textDocument/didSave. If the file is re-saved before the previous indexing has complete, the save will fail and often produce various ambiguous errors.

I don't use Windows and am not clear what Windows does. If my understanding is correct, we need a fix similar to https://github.com/MaskRay/ccls/commit/d4871207ed44c8e5edf0961ae7736c24075486ba. That commit fixed sema_manager.cc (was `clang_complete.cc`; it implements completion and diagnostics) but we probably need a fix in indexer.cc (the component that indexes a file. The index will serve various textDocument/{definition,references,etc} $ccls/member etc requests).

> I have been using this for a month and I believe clangd has something very similar implemented. No issues so far, even on very large source files, so I want to merge if possible.

If you can reliably reproduce this issue. Would you mind trying the following patch?
It should fix the same issue but the change is much less intrusive.

```patch
diff --git i/src/indexer.cc w/src/indexer.cc
index 280e3c4d..69eb7291 100644
--- i/src/indexer.cc
+++ w/src/indexer.cc
@@ -1263,6 +1263,8 @@ Index(SemaManager *manager, WorkingFiles *wfiles, VFS *vfs,
   Clang->setVirtualFileSystem(FS);
   Clang->createFileManager();
 #endif
+  Clang->setSourceManager(new SourceManager(Clang->getDiagnostics(),
+                                            Clang->getFileManager(), true));
 
   IndexParam param(*vfs, no_linkage);
   auto DataConsumer = std::make_shared<IndexDataConsumer>(param);
```

Crunkle:
That is great, thanks! I was looking at the best way to set `UserFilesAreVolatile` but totally overlooked reconstructing the source manager. I have now tested your solution today and it has completely resolved all of my issues. It's far more elegant than my previous commit so I agree, we should definitely use your patch instead.

As for precompiled headers: I did see that clangd makes an exception for them but I don't think there is any valid rationale for us to do the same. Your patch fixes write locks, especially after saving the file, and so precompiled shouldn't cause any problems.

I have updated my pull request but feel free to close and push whatever you think is best. You are more familiar with this than me but I will be very happy to see this resolved.

MaskRay:
@Crunkle Can you confirm that without the fix, you'll see issues like #235?

> As for precompiled headers: I did see that clangd makes an exception for them but I don't think there is any valid rationale for us to do the same. Your patch fixes write locks, especially after saving the file, and so precompiled shouldn't cause any problems.

Can you elaborate the exception?

Crunkle:
Yes, without the fix I have various issues: unable to open the file for writing. When using the fix I have no issues and the debugger shows all is working correctly with no memory maps.

> Can you elaborate the exception?

See clangd's [FSProvider.cpp:39](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clangd/FSProvider.cpp#L39). This is the check to skip generated premables as they are similarly building them themselves. That may be more optimal but I do not think it is necessary to do so.

Also note clang's [Basic/SourceManager.cpp:137](https://github.com/llvm-mirror/clang/blob/master/lib/Basic/SourceManager.cpp#L137).

PR #461, opened 2019-08-11
Support override CLANG_RESOURCE_DIR

zhsj:
The result from clang is something like
/usr/lib/llvm-7/lib/clang/7.0.1

Which means I need to recompile when the clang is upgraded to later 7.x.x

However the system has a convenient symlink, which is
/usr/include/clang/7.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,22 +120,24 @@ endif()
 
 # Find Clang resource directory with Clang executable.
 
-find_program(CLANG_EXECUTABLE clang)
-if(NOT CLANG_EXECUTABLE)
-  message(FATAL_ERROR "clang executable not found.")
-endif()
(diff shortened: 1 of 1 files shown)


MaskRay:
If `CLANG_RESOURCE_DIR` is specified, the following lines should not be executed:

```cmake
find_program(CLANG_EXECUTABLE clang)
if(NOT CLANG_EXECUTABLE)
  message(FATAL_ERROR "clang executable not found.")
endif()
```

zhsj:
> If `CLANG_RESOURCE_DIR` is specified, the following lines should not be executed:
> 
> ```cmake
> find_program(CLANG_EXECUTABLE clang)
> if(NOT CLANG_EXECUTABLE)
>   message(FATAL_ERROR "clang executable not found.")
> endif()
> ```

fixed

PR #467, opened 2019-08-17
Fix some clang-tidy performance warnings

DanielChabrowski:
Hello,
I run some clang-tidy performance-* checks on the code and decided to correct the warnings. Most of them warn about unnecessary copy or to preallocate a vector.

It seems that clang-format-8 added some additional unintended changes.

Cheers.

### DIFF
--- a/src/lsp.cc
+++ b/src/lsp.cc
@@ -63,7 +63,7 @@ void DocumentUri::setPath(const std::string &path) {
   // file:///c%3A/Users/jacob/Desktop/superindex/indexer/full_tests
   raw_uri = path;
 
-  size_t index = raw_uri.find(":");
+  size_t index = raw_uri.find(':');
   if (index == 1) { // widows drive letters must always be 1 char
     raw_uri.replace(raw_uri.begin() + index, raw_uri.begin() + index + 1,
                     "%3A");
--- a/src/message_handler.cc
+++ b/src/message_handler.cc
@@ -121,9 +121,8 @@ void ReplyOnce::replyLocationLink(std::vector<LocationLink> &result) {
   if (g_config->client.linkSupport) {
     (*this)(result);
   } else {
-    std::vector<Location> result1;
-    for (auto &loc : result)
-      result1.emplace_back(std::move(loc));
+    std::vector<Location> result1{std::make_move_iterator(result.begin()),
+                                  std::make_move_iterator(result.end())};
     (*this)(result1);
   }
 }
--- a/src/message_handler.cc
+++ b/src/message_handler.cc
@@ -121,8 +121,8 @@ void ReplyOnce::replyLocationLink(std::vector<LocationLink> &result) {
   if (g_config->client.linkSupport) {
     (*this)(result);
   } else {
-    std::vector<Location> result1{std::make_move_iterator(result.begin()),
-                                  std::make_move_iterator(result.end())};
+    std::vector<Location> result1(std::make_move_iterator(result.begin()),
+                                  std::make_move_iterator(result.end()));
     (*this)(result1);
   }
 }
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -616,7 +616,7 @@ void mainLoop() {
 
   SemaManager manager(
       &project, &wfiles,
-      [&](const std::string &path, std::vector<Diagnostic> diagnostics) {
+      [](const std::string &path, std::vector<Diagnostic> diagnostics) {
         PublishDiagnosticParam params;
         params.uri = DocumentUri::fromPath(path);
         params.diagnostics = std::move(diagnostics);
@@ -718,7 +718,7 @@ void standalone(const std::string &root) {
   VFS vfs;
   SemaManager manager(
       nullptr, nullptr,
-      [&](const std::string &, const std::vector<Diagnostic> &) {},
+      [](const std::string &, const std::vector<Diagnostic> &) {},
       [](const RequestId &id) {});
   IncludeComplete complete(&project);
 
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -616,13 +616,13 @@ void mainLoop() {
 
   SemaManager manager(
       &project, &wfiles,
-      [&](std::string path, std::vector<Diagnostic> diagnostics) {
+      [&](const std::string &path, std::vector<Diagnostic> diagnostics) {
         PublishDiagnosticParam params;
         params.uri = DocumentUri::fromPath(path);
-        params.diagnostics = diagnostics;
+        params.diagnostics = std::move(diagnostics);
         notify("textDocument/publishDiagnostics", params);
       },
-      [](RequestId id) {
+      [](const RequestId &id) {
         if (id.valid()) {
           ResponseError err;
           err.code = ErrorCode::InternalError;
@@ -717,8 +717,9 @@ void standalone(const std::string &root) {
   WorkingFiles wfiles;
   VFS vfs;
   SemaManager manager(
-      nullptr, nullptr, [&](std::string, std::vector<Diagnostic>) {},
-      [](RequestId id) {});
+      nullptr, nullptr,
+      [&](const std::string &, const std::vector<Diagnostic> &) {},
+      [](const RequestId &id) {});
   IncludeComplete complete(&project);
 
   MessageHandler handler;
@@ -756,7 +757,7 @@ void standalone(const std::string &root) {
 void index(const std::string &path, const std::vector<const char *> &args,
            IndexMode mode, bool must_exist, RequestId id) {
   pending_index_requests++;
-  index_request->pushBack({path, args, mode, must_exist, id},
+  index_request->pushBack({path, args, mode, must_exist, std::move(id)},
                           mode != IndexMode::Background);
 }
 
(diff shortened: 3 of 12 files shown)


src/clang_tu.cc
```diff
@@ -104,7 +104,7 @@ BuildCompilerInvocation(const std::string &main, std::vector<const char *> args,
       CompilerInstance::createDiagnostics(new DiagnosticOptions,
                                           new IgnoringDiagConsumer, true));
   std::unique_ptr<CompilerInvocation> CI =
-      createInvocationFromCommandLine(args, Diags, VFS);
+      createInvocationFromCommandLine(args, Diags, std::move(VFS));
```
MaskRay: This is not very necessary because VFS itself is just a ref counter..

src/lsp.cc
```diff
@@ -63,7 +63,7 @@ void DocumentUri::SetPath(const std::string &path) {
   // file:///c%3A/Users/jacob/Desktop/superindex/indexer/full_tests
   raw_uri = path;
 
-  size_t index = raw_uri.find(":");
+  size_t index = raw_uri.find(':');
```
MaskRay: This is fine

src/message_handler.cc
```diff
@@ -228,7 +227,8 @@ void MessageHandler::Run(InMessage &msg) {
       } catch (NotIndexed &) {
         throw;
       } catch (...) {
-        reply.Error(ErrorCode::InternalError, "failed to process " + msg.method);
+        reply.Error(ErrorCode::InternalError,
```
MaskRay: These intermixed formatting is not very necessary..

src/message_handler.cc
```diff
@@ -121,9 +121,8 @@ void ReplyOnce::ReplyLocationLink(std::vector<LocationLink> &result) {
   if (g_config->client.linkSupport) {
     (*this)(result);
   } else {
-    std::vector<Location> result1;
-    for (auto &loc : result)
-      result1.emplace_back(std::move(loc));
+    std::vector<Location> result1{std::make_move_iterator(result.begin()),
```
MaskRay: This is good.

src/project.hh
```diff
@@ -76,7 +76,7 @@ struct Project {
   void SetArgsForFile(const std::vector<const char *> &args,
                       const std::string &path);
 
-  void Index(WorkingFiles *wfiles, RequestId id);
+  void Index(WorkingFiles *wfiles, const RequestId &id);
```
MaskRay: This is fallout of https://github.com/MaskRay/ccls/issues/437 ..

src/query.cc
```diff
@@ -357,7 +357,7 @@ void DB::ApplyIndexUpdate(IndexUpdate *u) {
       RefDecl(prev_lid2file_id, usr, Kind::Func, *def.spell, -1);
   RemoveUsrs(Kind::Func, u->file_id, u->funcs_removed);
   Update(lid2file_id, u->file_id, std::move(u->funcs_def_update));
-  for (auto &[usr, del_add]: u->funcs_declarations) {
+  for (auto &[usr, del_add] : u->funcs_declarations) {
```
MaskRay: Unnecessary formatting changes

src/query.cc
```diff
@@ -476,7 +476,7 @@ void DB::Update(const Lid2file_id &lid2file_id, int file_id,
 void DB::Update(const Lid2file_id &lid2file_id, int file_id,
                 std::vector<std::pair<Usr, QueryVar::Def>> &&us) {
   for (auto &u : us) {
-    auto &def = u.second;
+    auto def = u.second;
```
MaskRay: This is wrong.

src/sema_manager.cc
```diff
@@ -245,14 +245,13 @@ class StoreDiags : public DiagnosticConsumer {
   void BeginSourceFile(const LangOptions &Opts, const Preprocessor *) override {
     LangOpts = &Opts;
   }
-  void EndSourceFile() override {
```
MaskRay: Unnecessary formatting.

src/serializer.cc
```diff
@@ -35,7 +35,7 @@ bool gTestOutputMode = false;
 
 namespace ccls {
 
-void JsonReader::IterArray(std::function<void()> fn) {
+void JsonReader::IterArray(const std::function<void()> &fn) {
```
MaskRay: I would change this to use function_ref instead

src/test.cc
```diff
@@ -15,16 +15,16 @@ limitations under the License.
```
MaskRay: test.cc is under maintenance and I'd probably remove it. So further changes to this file is not needed.

src/message_handler.cc
DanielChabrowski: The project isn't formatted properly therefore the changes. I can try to remove formatting from this PR and make another one with pure clang-format run on src directory. Would that be okay?

MaskRay:
Since there are many accumulated formatting problems and I'd like to change variable/function cases, I'll do an enormous formatting change: https://github.com/MaskRay/ccls/pull/471

DanielChabrowski:
> Since there are many accumulated formatting problems and I'd like to change variable/function cases, I'll do an enormous formatting change: #471

Did you update .clang-format config? It still modifies pretty much every file for me. Tried both clang-format 7 and 8.

MaskRay:
> > Since there are many accumulated formatting problems and I'd like to change variable/function cases, I'll do an enormous formatting change: #471
> 
> Did you update .clang-format config? It still modifies pretty much every file for me. Tried both clang-format 7 and 8.

I ran `git diff -U0 --no-color 'HEAD^' | ~/llvm/tools/clang/tools/clang-format/clang-format-diff.py -i -p1` to format lines touched by the variable/function renaming change.
I am using clang-format built from trunk. It may be different a bit from clang-format 7 and 8.

XVilka:
There are a lot of conflicts now that should be resolved.

src/serializer.hh
```diff
@@ -48,8 +49,8 @@ struct JsonReader {
   JsonReader(rapidjson::Value *m) : m(m) {}
   void startObject() {}
   void endObject() {}
-  void iterArray(std::function<void()> fn);
-  void member(const char *name, std::function<void()> fn);
+  void iterArray(const llvm::function_ref<void()> fn);
```
MaskRay: Remove `const `

src/serializer.cc
```diff
@@ -48,7 +48,7 @@ void JsonReader::iterArray(std::function<void()> fn) {
   }
   path_.pop_back();
 }
-void JsonReader::member(const char *name, std::function<void()> fn) {
+void JsonReader::member(const char *name, const llvm::function_ref<void()> fn) {
```
MaskRay: Remove `const `

src/serializer.cc
```diff
@@ -35,7 +35,7 @@ bool gTestOutputMode = false;
 
 namespace ccls {
 
-void JsonReader::iterArray(std::function<void()> fn) {
+void JsonReader::iterArray(const llvm::function_ref<void()> fn) {
```
MaskRay: Remove `const `

src/sema_manager.cc
```diff
@@ -231,7 +231,7 @@ class StoreDiags : public DiagnosticConsumer {
   }
 
 public:
-  StoreDiags(std::string path) : path(path) {}
+  StoreDiags(std::string path) : path{std::move(path)} {}
```
MaskRay: `path(std::move(path))`

src/pipeline.cc
```diff
@@ -717,8 +717,9 @@ void standalone(const std::string &root) {
   WorkingFiles wfiles;
   VFS vfs;
   SemaManager manager(
-      nullptr, nullptr, [&](std::string, std::vector<Diagnostic>) {},
-      [](RequestId id) {});
+      nullptr, nullptr,
+      [&](const std::string &, const std::vector<Diagnostic> &) {},
```
MaskRay: Delete `&`

src/message_handler.cc
MaskRay: Use `(..)` instead of `{...}`. `{}` can be confusing because the constructor has an overload on initializer_list.

src/pipeline.cc
```diff
@@ -616,13 +616,13 @@ void mainLoop() {
 
   SemaManager manager(
       &project, &wfiles,
-      [&](std::string path, std::vector<Diagnostic> diagnostics) {
+      [&](const std::string &path, std::vector<Diagnostic> diagnostics) {
```
MaskRay: `[&]` -> `[]`

MaskRay: sema_manager.hh `OnDiagnostic` can be changed to llvm::function_ref now.

DanielChabrowski: Lambda capture done. OnDiagnostic & OnDropped part can be delivered by a different PR.

MaskRay:
Thanks a lot! (and bear with me making it hard to rebase). I'll make some more adjustments and merge.

The below is a false positive.

```cpp
src/utils.cc:155:7: warning: string concatenation results in allocation of unnecessary temporary strings; consider using 'operator+=' or 'string::append()' instead [performance-inefficient-string-concatenation]
      path = root + path.substr(real.size());
```

Changing it to `string::append` will destroy `root`.

`serializer.hh` should not include too many llvm headers. `#include <llvm/ADT/STLExtras.h>` should be moved to `serializer.cc`

PR #475, opened 2019-08-30
Fix build errors on Windows after massive rename.

Riatre:


MaskRay:
Thanks! Merged with the massive rename.

PR #484, opened 2019-09-17
indexer: decode invalid UTF-8 byte sequence in comments as ?

MaskRay:


src/indexer.cc
```diff
@@ -456,7 +457,24 @@ class IndexDataConsumer : public index::IndexDataConsumer {
                (*p == ' ' || *p == '/' || *p == '*' || *p == '<' || *p == '!'))
           prefix--, p++;
       }
-      ret.insert(ret.end(), p, q);
+      if (high) {
+        while (p < q) {
+          int i = 0, c = *p < 0x80 ? 0
+                                   : *p < 0xc0 || *p >= 0xf8
```
chenz: Should this not be `*p < 0xc2 || *p >= 0xf5`?
Also I think the continuation bytes loop should check for `*r >= 0x80 && *r < 0xc0`

Reference: https://en.wikipedia.org/wiki/UTF-8#Codepage_layout

MaskRay: > *r >= 0x80 && *r < 0xc0

Omitting `*r < 0xc0` is a compromise to decode `A B` as a single `?` if `0xc2<=A<0xf5 && B>=0xc0`

PR #518, opened 2019-11-01
Fix builds for LLVM 9 (for LLVM build without BUILD_SHARED_LIBS=ON)

dcermak:
LLVM 9 officially ships the C++ libraries as a single library `libclang-cpp.so`. This makes ccls fail to link as the individual libraries `libclangIndex.so`, `libclangFormat.so`, etc. are no longer present.

We instead try to find the clang-cpp library and if it is found we link against it instead of the individual libraries. If it is not found, then we proceed as previously.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,23 +68,33 @@ endif()
 
 ### Libraries
 
+# LLVM >= 9 ships the C++ libraries by default in a single library libclang-cpp,
+# but this is not universally enabled on all distributions.
+# If clang-cpp is available, then link against it, otherwise link against the
+# individual clang libraries.
 find_package(Clang REQUIRED)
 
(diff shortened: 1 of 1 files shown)


MaskRay:
Thanks for reminding me of this option.

> LLVM 9 officially ships the C++ libraries as a single library libclang-cpp.so. This makes ccls fail to link as the individual libraries libclangIndex.so, libclangFormat.so, etc. are no longer present.

LLVM 9 supports CLANG_LINK_CLANG_DYLIB. The prebuilt binaries downloaded from releases.llvm.org/9.0.0 include `libclang*.a` and `libclang-cpp.so`. You can choose between 2 linking methods, see https://github.com/MaskRay/ccls/pull/519

Issue #13, opened 2018-05-14
link error to LLVMSupport


scturtle:
```
CMakeFiles/ccls.dir/src/main.cc.o: In function `OptionValue':
/home/shc/code/repos/ccls/build/clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-16.04/include/llvm/Support/CommandLine.h:575: undefined reference to `vtable for llvm::cl::OptionValue<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >'
CMakeFiles/ccls.dir/src/main.cc.o: In function `opt<char [5], llvm::cl::desc>':
/home/shc/code/repos/ccls/build/clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-16.04/include/llvm/Support/CommandLine.h:1359: undefined reference to `vtable for llvm::cl::opt<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, false, llvm::cl::parser<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >'
CMakeFiles/ccls.dir/src/main.cc.o: In function `parser':
/home/shc/code/repos/ccls/build/clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-16.04/include/llvm/Support/CommandLine.h:1028: undefined reference to `vtable for llvm::cl::parser<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >'
CMakeFiles/ccls.dir/src/main.cc.o: In function `OptionValue':
/home/shc/code/repos/ccls/build/clang+llvm-6.0.0-x86_64-linux-gnu-ubuntu-16.04/include/llvm/Support/CommandLine.h:575: undefined reference to `vtable for llvm::cl::OptionValue<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >'
CMakeFiles/ccls.dir/src/main.cc.o: In function `opt<char [9], llvm::cl::desc, llvm::cl::value_desc>':
[...]

PR #519, opened 2019-11-01
cmake: support CLANG_LINK_CLANG_DYLIB

MaskRay:
------------

Prebuilt binaries downloaded from http://releases.llvm.org/9.0.0/

Link against `libLLVM*.a libclang*.a`
`cmake -GNinja -H. -BBundled9C -DCMAKE_PREFIX_PATH=$PWD/build/clang+llvm-9.0.0-x86_64-linux-gnu-ubuntu-16.04`

```
% stat -c %s Bundled9/ccls
32432992
```

Link against `libLLVM*.a libclang-cpp.so`
`cmake -GNinja -H. -BBundled9C -DCMAKE_PREFIX_PATH=$PWD/build/clang+llvm-9.0.0-x86_64-linux-gnu-ubuntu-16.04 -DCLANG_LINK_CLANG_DYLIB=on`

```
% stat -c %s Bundled9C/ccls
1774992
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,21 +70,25 @@ endif()
 
 find_package(Clang REQUIRED)
 
-target_link_libraries(ccls PRIVATE
-  clangIndex
-  clangFormat
-  clangTooling
-  clangToolingInclusions
-  clangToolingCore
-  clangFrontend
-  clangParse
-  clangSerialization
-  clangSema
-  clangAST
-  clangLex
-  clangDriver
-  clangBasic
-)
+if(CLANG_LINK_CLANG_DYLIB)
+  target_link_libraries(ccls PRIVATE clang-cpp)
+else()
+  target_link_libraries(ccls PRIVATE
+    clangIndex
+    clangFormat
+    clangTooling
+    clangToolingInclusions
+    clangToolingCore
+    clangFrontend
+    clangParse
+    clangSerialization
+    clangSema
+    clangAST
+    clangLex
+    clangDriver
+    clangBasic
+  )
+endif()
 
 if(LLVM_LINK_LLVM_DYLIB)
   target_link_libraries(ccls PRIVATE LLVM)


dcermak:
This works when building ccls on openSUSE Tumbleweed with llvm9 from the openSUSE repositories, **but** the user must explicitly set `-DCLANG_LINK_CLANG_DYLIB=on` in the cmake call, otherwise they still get the linking error. The advantage of #518 is that it works "out of the box" and doesn't require additional flags to be passed to cmake.

MaskRay:
> This works when building ccls on openSUSE Tumbleweed with llvm9 from the openSUSE repositories, **but** the user must explicitly set `-DCLANG_LINK_CLANG_DYLIB=on` in the cmake call, otherwise they still get the linking error. The advantage of #518 is that it works "out of the box" and doesn't require additional flags to be passed to cmake.

It is difficult to know which of `libclang*.a` and `libclang-cpp.so` is available. They may both exist (e.g. `clang+llvm-9.0.0-x86_64-linux-gnu-ubuntu-16.04`). For that case, I want the users to have freedom to choose the linking method, as documented in my examples.

dcermak:
Well, your project, your call.

Nevertheless, I'd like to suggest that the default (i.e. `cmake .. && make`) should just work, which is not the case with this PR (it fails to link on openSUSE with LLVM9 from the openSUSE repositories).

MaskRay:
Let me reiterate, `cmake -Hllvm -BRelease && make -C Release` just works for a standard llvm+clang build. I suspect the openSUSE package provides libclang-cpp.so but not `libclang*.a libclang*.so`. This is understandable because llvm libraries are huge. `libclang-cpp.so` provides for lots of scenarios. For example, on Arch Linux, `extra/llvm` and `extra/clang` packages are somewhat non-standard and it can be tricky to use for a dependent package, see https://bugs.archlinux.org/task/60512 .

Using `libclang-cpp.so` is not always desirable. I have seen issues like an LLVM global constructor (llvm::opt::Option) registered twice (e.g. https://github.com/MaskRay/ccls/issues/13#issuecomment-388690797).

PR #521, opened 2019-11-08
Improve completion cache.

iensen:
Unfortunately, this change https://github.com/MaskRay/ccls/commit/1a41d2dbd18b6548383b625b960b9178acdb144d made auto-completion for my large projects completely unusable.

My scenario is as follows.

Suppose I have a class variable c, and I want to type c->some_method.
When I type c->, I wait 0.5-1 second until some_method appears. (my idle-delay is set to 0, but it doesn't really matter).

Then, when I type o (the next character), emacs freezes and for 0.5 -1 seconds before it generates new completions request. This happens because it no longer uses the cache and the completions are recomputed again.  
This is super inconvenient! Previously, we used to reuse the cache for position c-> and filter it out when the new characters are added.

As far as I understand, my change should fix the bug which the original commit was supposed to fix, and will satisfy the important use case I desribed above (I verified the latter). But I may be misunderstanding the code, let me know.

### DIFF
--- a/src/sema_manager.hh
+++ b/src/sema_manager.hh
@@ -176,7 +176,7 @@ template <typename T> struct CompleteConsumerCache {
                     Position position) {
     std::lock_guard lock(mutex);
     return this->path == path && this->position == position &&
-           this->line == line;
+            line.substr(0,position.character) == this->line.substr(0, position.character);
   }
 };
 } // namespace ccls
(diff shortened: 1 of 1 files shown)


src/sema_manager.hh
```diff
@@ -176,7 +176,7 @@ template <typename T> struct CompleteConsumerCache {
                     Position position) {
     std::lock_guard lock(mutex);
     return this->path == path && this->position == position &&
-           this->line == line;
+            line.substr(0,position.character) == this->line.substr(0, position.character);
```
MaskRay: space after `,`

iensen: Cool, thanks.

PR #522, opened 2019-11-11
initialize: Implement serverInfo (3.15.0)

nemethf:
I don't really like the repeated set_property in CMakeList.txt, but wasn't able to find an easy, alternative way to access the version info in initialize.cc.

Thanks.

----

Return serverInfo for an initialize request as the upcoming version of
LSP specifies.  This helps clients to identify ccls even when they
just connect to a TCP port and allows them to easily implement server
specific functionalities like $ccls/navigate.


MaskRay:
Thanks! Tweaked a bit and merged.

PR #524, opened 2019-11-15
Add name for anonymous Enum

xVanTuring:
https://github.com/MaskRay/ccls/blob/bd609e89a29c508c8c763db2ecfad50e207391b3/src/indexer.cc#L963-L965
Here, `case Decl::Record` is trying to match `enum`.  However, `EnumDecl` and `RecordDecl` are both inherited from [`TagDecl`](https://clang.llvm.org/doxygen/classclang_1_1TagDecl.html#details) directly, is, so the matching here will never get executed, which causes the anonymous Enum having no name. So I add a `case:: Enum` matching above `case Decl::Record`, and it will add a simple name for anonymous Enum.  

Of course, removing the matching code is an alternative solution.
Any edits are welcome.

### DIFF
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -874,10 +874,10 @@ public:
                 Usr usr1 = getUsr(d1, &info1);
                 IndexType &type1 = db->toType(usr1);
                 SourceLocation sl1 = d1->getLocation();
-                type1.def.spell = {
-                    Use{{fromTokenRange(sm, lang, {sl1, sl1}), Role::Definition},
-                        lid},
-                    fromTokenRange(sm, lang, sr1)};
+                type1.def.spell = {Use{{fromTokenRange(sm, lang, {sl1, sl1}),
+                                        Role::Definition},
+                                       lid},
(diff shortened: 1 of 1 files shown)


MaskRay:
Thanks. This will assign `anon enum` to an anonymous Enum.

```cpp
struct A {
  enum {B,C} xx;
};
```

textDocument/hover on `enum` will say "anon enum".

src/indexer.cc
```diff
@@ -874,10 +874,10 @@ class IndexDataConsumer : public index::IndexDataConsumer {
                 Usr usr1 = getUsr(d1, &info1);
                 IndexType &type1 = db->toType(usr1);
                 SourceLocation sl1 = d1->getLocation();
-                type1.def.spell = {
-                    Use{{fromTokenRange(sm, lang, {sl1, sl1}), Role::Definition},
-                        lid},
-                    fromTokenRange(sm, lang, sr1)};
+                type1.def.spell = {Use{{fromTokenRange(sm, lang, {sl1, sl1}),
```
MaskRay: I deleted this unrelated change when merging.

PR #525, opened 2019-11-18
Support the 'reference' codeActionKind

nemethf:
I wasn't sure whether my idea of supporting extra reference methods with a slight extension of the language server protocol would work, so I implemented this proof-of-concept.  Unfortunately, it turned out that I don't really know c++.  On the other hand, the code works.   The corresponding code implementing the client side is here: https://github.com/nemethf/eglot-x/tree/reference-codeActionKind

And this is the summary of the 'reference' codeActionKind extension:
https://github.com/joaotavora/eglot/issues/302#issuecomment-550225329

I hope you find this useful, and I'm eager to receive your feedback.


MaskRay:
Thanks for working on this. My feeling regarding https://github.com/joaotavora/eglot/issues/302#issuecomment-550225329 is that to make less use of extension methods, it adds some out-of-place round trips. It does reduce the number of methods, but is it really worthwhile? There are still some extensions to CodeAction anyway.

nemethf:
My aim is to come up with a server-agnostic approach.  Currently if clangd implements the same $ccls/* methods, then the clients need to be modified to add special cases for the clangd server as well. 

It is true that the reference actionKind  approach does not expose all the capabilities of the $ccls methods, but it provides a useful subset.   Clients can support additional features with server-specific code.  I think the extra round-trip does not result in noticeable extra delay, because the clients should wait for user input between the two method calls.   

An alternative approach is to list the names of the extra reference methods among the server's capabilities.  However, this approach does not support additional arguments.  Would you accept this?

> is it really worthwhile? 

I looked at a bunch of LSP servers.  It seems ccls is the only one that has this kind of protocol extensions.  So, at the moment, it's more like a theoretical change than a practical one.

PR #538, opened 2019-12-03
set background thread priority on index threads

eklitzke:
Previously I created #537 which uses the Linux-specific `SCHED_BATCH`, but then I noticed that LLVM has a platform-agnostic way of doing this that works on Windows, macOS, etc. Note that the LLVM implementation uses `SCHED_IDLE` on Linux which is much more extreme than `SCHED_BATCH`, but probably still OK for the indexer 🤷

### DIFF
--- a/src/messages/initialize.cc
+++ b/src/messages/initialize.cc
@@ -12,6 +12,7 @@
 #include "working_files.hh"
 
 #include <llvm/ADT/Twine.h>
+#include <llvm/Config/llvm-config.h>
 #include <llvm/Support/Threading.h>
 
 #include <rapidjson/document.h>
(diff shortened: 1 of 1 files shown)


MaskRay:
It is not very clear what `pthread_setschedparam(pthread_self(), SCHED_IDLE, ...)` will achieve. Have you noticed any difference with this setting?

eklitzke:
> It is not very clear what `pthread_setschedparam(pthread_self(), SCHED_IDLE, ...)` will achieve. Have you noticed any difference with this setting?

Moved discussion to #536 as these are similar.

src/messages/initialize.cc
```diff
@@ -255,6 +255,9 @@ void *indexer(void *arg_) {
   delete arg;
   std::string name = "indexer" + std::to_string(idx);
   set_thread_name(name.c_str());
+#if defined(LLVM_ENABLE_THREADS) && LLVM_VERSION_MAJOR >= 9
```
MaskRay: `#if LLVM_ENABLE_THREADS && LLVM_VERSION_MAJOR >= 9` 

[`llvm/include/llvm/Config/llvm-config.h.cmake`](https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/Config/llvm-config.h.cmake) says
says `#cmakedefine01 LLVM_ENABLE_THREADS`

PR #548, opened 2019-12-22
textDocument/documentSymbol should work for non-opened files as well.

jlahoda:
I was trying to use ccls as a C/C++ server for the Apache NetBeans IDE, and it turned out the textDocument/documentSymbol only works for opened files. That is unfortunate, as NetBeans is using this request to fill the "Navigator", which generally works for any selected file, even those that are just selected in one of the files view. Moreover, the LSP specification says that: "Note that a server’s ability to fulfill requests is independent of whether a text document is open or closed.", so it would seem reasonable to me if the server would answer the query even for non-opened files. In the case of textDocument/documentSymbol, it seems this may be reasonably doable, see this patch.

What do you think?

### DIFF
--- a/src/message_handler.cc
+++ b/src/message_handler.cc
@@ -252,9 +252,9 @@ QueryFile *MessageHandler::findFile(const std::string &path, int *out_file_id) {
 
 std::pair<QueryFile *, WorkingFile *>
 MessageHandler::findOrFail(const std::string &path, ReplyOnce &reply,
-                           int *out_file_id) {
+                           int *out_file_id, bool allow_unopened) {
   WorkingFile *wf = wfiles->getFile(path);
-  if (!wf) {
+  if (!wf && !allow_unopened) {
     reply.notOpened(path);
     return {nullptr, nullptr};
   }
--- a/src/message_handler.hh
+++ b/src/message_handler.hh
@@ -236,7 +236,8 @@ struct MessageHandler {
   QueryFile *findFile(const std::string &path, int *out_file_id = nullptr);
   std::pair<QueryFile *, WorkingFile *> findOrFail(const std::string &path,
                                                    ReplyOnce &reply,
-                                                   int *out_file_id = nullptr);
+                                                   int *out_file_id = nullptr,
+                                                   bool allow_unopened = false);
 
 private:
   void bind(const char *method, void (MessageHandler::*handler)(JsonReader &));
--- a/src/messages/textDocument_document.cc
+++ b/src/messages/textDocument_document.cc
@@ -147,8 +147,8 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
 
   int file_id;
   auto [file, wf] =
-      findOrFail(param.textDocument.uri.getPath(), reply, &file_id);
-  if (!wf)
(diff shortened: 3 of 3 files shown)


MaskRay:
This probably works for `textDocument/documentSymbol` because it only uses `WorkingFile` in `getLsRange(wf, sym.range)`, which does not require `wf` to be non-null. If the document is not opened, `textDocument/documentLink`, `textDocument/references`, `textDocument/definitions` may trigger a null pointer dereference.

The specification does say

> Note that a server’s ability to fulfill requests is independent of whether a text document is open or closed.

But I feel this can clutter server implementations and a client can actually circumvent easily by sending `textDocument/didOpen` first.

jlahoda:
Right, this may not work so easily for other queries (although I am not sure why it should be impossible or prohibitively difficult to implement them). But I so far do not have a strong usecase for other queries (textDocument/references, could admittedly be useful when invoked through the Navigator, though). Hence this proposed patch only changes textDocument/documentSymbol, and nothing else, I believe.

And right again that the clients could workaround by doing didOpen/didClose. But, besides cluttering client implementations (just for the sake of a very limited set of servers), this also brings a question of how many files should be kept virtually open, when to virtually close them, etc. And this all feels like a huge waste of resources - presumably when didOpen is issued, the server will start to compute diagnostics as if the file was really opened, but the only thing that is really requested is the document's structure, which (per my understanding) is available, and does not really need to be computed!

Overall, I think the protocol puts some constraints on the clients, and some on the servers. And some of them are difficult to achieve (and this is true for both clients and servers). But I am not convinced that ignoring some statements of the spec is a good way to make the clients and server cooperate for the benefit of the users. All I am trying to do is to fix the problem where I think they are (and there were multiple problems on the client side exposed by ccls, which could be workarounded in ccls - but that would be a wrong place to do that!)

src/message_handler.cc
```diff
@@ -252,9 +252,10 @@ QueryFile *MessageHandler::findFile(const std::string &path, int *out_file_id) {
 
 std::pair<QueryFile *, WorkingFile *>
 MessageHandler::findOrFail(const std::string &path, ReplyOnce &reply,
-                           int *out_file_id) {
+                           int *out_file_id,
+                           const bool acceptNotOpened) {
```
MaskRay: `bool allow_unopened`

src/message_handler.hh
```diff
@@ -236,7 +236,8 @@ struct MessageHandler {
   QueryFile *findFile(const std::string &path, int *out_file_id = nullptr);
   std::pair<QueryFile *, WorkingFile *> findOrFail(const std::string &path,
                                                    ReplyOnce &reply,
-                                                   int *out_file_id = nullptr);
+                                                   int *out_file_id = nullptr,
+                                                   const bool acceptNotOpened = false);
```
MaskRay: `bool allow_unopened = false`

jlahoda:
Sure, here is a screenshot:
navigator-c

The Navigator is in the bottom left corner - the content changes based on what is selected in the "Projects" view above (or in several other views, or in editor). No editor needs to be opened at all - it is just the file's structure. textDocument/documentLink (and most others) is not really needed, I think, as there is not a good way to show that to the user anyway.

jlahoda:
Thank you!

PR #560, opened 2020-01-29
Convert llvm::StringRef -> std::string from implicit to explicit

lanza:
A change in llvm 11 changed the llvm::StringRef API to disable implicit
casts from llvm::StringRef to std::string. Convert all calls here to
the explicit API.


MaskRay:
Thanks!

I was doing some linker changes and investigating an x86-64 `leaq sym+large_offset(%rip)` [bug](https://reviews.llvm.org/D73606) yesterday. I did not notice the breakage in time. I'll go with a different fix #561

MaskRay:
I'll also wait a bit for https://reviews.llvm.org/D73640

PR #561, opened 2020-01-29
Adapt llvmorg-11-init-1314-g777180a32b6: StringRef's conversion to std::string is now explicit.

MaskRay:


MaskRay:
Wait for https://reviews.llvm.org/D73640

PR #571, opened 2020-02-10
Simplify: hide unrelated options

condy0919:


MaskRay:
There is actually a difference, but it does not matter.

```patch
 OVERVIEW: C/C++/Objective-C language server
 
 See more on https://github.com/MaskRay/ccls/wiki
 USAGE: ccls [options]
 
 OPTIONS:
 
 Generic Options:
 
+  --help                - Display available options (--help-hidden for more)
   --help-list           - Display list of available options (--help-list-hidden for more)
+  --version             - Display the version of this program
 
 ccls options:
 
   -h                    - Alias for -help
   --index=<root>        - standalone mode: index a project and exit
   --init=<string>       - extra initialization options in JSON
   --log-file=<file>     - stderr or log file
   --log-file-append     - append to log file
   --test-index=<string> - run index tests
   -v=<int>              - verbosity, from -3 (fatal) to 2 (verbose)
```

Issue #580, opened 2020-02-25
ccls is failing to store a file from a sibling directory to the current root


lanza:
Given a project layout made via
```
git clone https://github.com/llvm/llvm-project
cd llvm-project
cmake -G Ninja -Sllvm -Bbuild/Debug -DLLVM_ENABLE_PROJECTS='clang'
cd clang
ln -s ../build/Debug/compile_commands.json .
```

and a rootPattern of `compile_commands.json` (thus ccls detects the root as `llvm-project/clang`) and a cache directory of `./.ccls-cache` (e.g. `llvm-project/llvm/.ccls-cache`) I consistently get this type of error:

```
utils.cc:173 E failed to write to /Users/lanza/Projects/llvm-project/clang/.ccls-cache/@Users@lanza@Projects@llvm-project@llvm/lib@Support@BlockFrequency.cpp No such file or directory
```

ccls is attempting to store files at `Users/lanza/Projects/llvm-project/clang/.ccls-cache/@Users@lanza@Projects@llvm-project@llvm/` though ccls has never made the directory. Simply `mkdir thedir`ing fixes the problem.

Also, while I'm at it, what's your personal set up for using ccls with llvm? I haven't found a configuration that doesn't surprise me with issues or necessary reindexing more often than I like.

MaskRay:
I don't recommend using a root pattern of `compile_commands.json`, because the client may locate `Debug/compile_commands.json` when visiting `Debug/include/llvm/Config/config.h`.

```
cd llvm-project
cmake -Hllvm -BDebug -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=on # other options
ln -s Debug/compile_commands.json .
touch .ccls-root
```

Configure the client to use `.ccls-root`.


(Slightly out of date: https://github.com/MaskRay/Config/wiki#llvm)

lanza:
Thanks for the setup! However, this doesn't solve the above fundamental problem. With your setup I still can hit, e.g.,

`16:07:13                     utils.cc:173 E failed to write to /Users/lanza/Projects/llvm-project/clang/.ccls-cache/@@Users@lanza@Projects@llvm-project@clang/@Applications@Xcode.app@Contents@Developer@Platforms@MacOSX.platform@Developer@SDKs@MacOSX.sdk@usr@include@_types@_wctrans_t.h No such file or directory
`

PR #583, opened 2020-02-28
Make a workspace dir before trying to save to is

lanza:
When ccls finds additional workspaceFolders it'll try to save indexes
in paths defined by mainWorkspaceDir/additionalWorkspaceDir/filepath
but it never attempts to create the additional dir. Fix that here.

### DIFF
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -196,7 +196,14 @@ Kind getKind(const Decl *d, SymbolKind &kind) {
     // ccls extension
     kind = SymbolKind::Parameter;
     return Kind::Var;
-  case Decl::Var:
+  case Decl::Var: {
+    auto vd = cast<VarDecl>(d);
+    if (vd->isStaticDataMember()) {
+      kind = SymbolKind::Field;
+      return Kind::Var;
+    }
+    [[fallthrough]];
+  }
   case Decl::Decomposition:
     kind = SymbolKind::Variable;
     return Kind::Var;


lanza:
Fixes #580

MaskRay:
Are you using workspace folders and the workspace folders are not passed as params in the `initialize` request?

```
   400          } else {
   401            if (g_config->cache.hierarchicalPath)
   402              sys::fs::create_directories(
   403                  sys::path::parent_path(cache_path, sys::path::Style::posix),
   404                  true);
   405            writeToFile(cache_path, curr->file_contents);
   406            writeToFile(appendSerializationFormat(cache_path),
   407                        serialize(g_config->cache.format, *curr));
   408          }
   409        }
```

The directories should have been created around `src/messages/initialize.cc:370`

lanza:
> Are you using workspace folders and the workspace folders are not passed as params in the initialize request?

No, ccls is just finding that files from other directories. It'll find a file, parse it normally and find that  the directory it wishes to save the index to doesn't exist. The log will contain `failed to write`.

MaskRay:
> > Are you using workspace folders and the workspace folders are not passed as params in the initialize request?
> 
> No, ccls is just finding that from other directories. (e.g. editing ccls files with a compile_commands.json built against `src/llvm-project/build/Release`.) It'll find a file, parse it normally and find that the directory it wishes to save the index to doesn't exist. The log will contain `failed to write`.

What does "other directories" mean? Can you give an example?

The project root
How it finds a different `compile_commands.json` (this is wrong unless "workspace folders" is used)
Which file causes `getCachePath` to return a file whose directory is not existent.

PR #587, opened 2020-03-11
Treat function-scope static variables as locals

### LINKED ISSUE #77: Wrong suggestion with const member function.

acowley:
If we report such variables as symbols in response to a
`textDocument/documentSymbol`, then we can no longer navigate to the
containing function using imenu in emacs.

Fixes [#77](https://github.com/MaskRay/emacs-ccls/issues/77)

### DIFF
--- a/src/indexer.hh
+++ b/src/indexer.hh
@@ -247,7 +247,7 @@ struct VarDef : NameMixin<VarDef> {
             parent_kind == SymbolKind::StaticMethod ||
             parent_kind == SymbolKind::Constructor) &&
            (storage == clang::SC_None || storage == clang::SC_Auto ||
-            storage == clang::SC_Register);
+            storage == clang::SC_Register || storage == clang::SC_Static);
   }
 
   const Usr *bases_begin() const { return nullptr; }


MaskRay:
From https://github.com/emacs-lsp/lsp-mode/issues/1424

> Ask a question in emacs-devel and file a bug there is no option for that.
> Ask @MaskRay to provide a setting in ccls to ignore variables in function bodies.
> Use alternative that works - lsp-treemacs-symbols works fine.
> File a feature request for helm-lsp to implement imenu like interface without losing this info.

Maybe we should report it to emacs-devel. A function-local static variable has static duration duration . Such variables may be interesting in the textDocument/documentSymbol result.

acowley:
I don't relish the idea of losing the duration information like this, but my concern is that this will be deemed out of scope for imenu. I use imenu for this because I like using it with helm, but the suggestion to use `lsp-treemacs-symbols` is perfectly valid.

This limitation of imenu is apparent elsewhere, so I think it is expected behavior. In an org-mode document, for example, `imenu` presents top-level headings. You can select one, and see the sub-headings below it, but you can not (afaict) navigate to that top-level heading. It seems like a limitation of the available interaction: you hit `enter` on an imenu item and it either drills down to sub-items, or moves point to that item if it is a leaf.

I'm okay with closing this PR as it's no trouble for me to continue using the patch, and I agree that in a perfect world this wouldn't be a change imposed on ccls.

Issue #594, opened 2020-03-21
The .gitignore does not match build instructions


php1ic:
### Observed behavior

Following the [build instructions](https://github.com/MaskRay/ccls/wiki/Build) does not leave the repo in a clean state.

### Expected behavior

A build directory named the same as in the instructions should be ignored by git.

### Steps to reproduce

1. Follow the [build instructions](https://github.com/MaskRay/ccls/wiki/Build) to create a Release build.

### System information

* ccls version (`git describe --tags --long`): 0.20190823.5-19-g73e8a224
* clang version: Any 
* OS: Any
* Editor: n/a
* Language client (and version): n/a

MaskRay:
Fixed by db12223081767fac2826512521bdeb61b6c01e88 (use a .gitignore whitelist)

PR #595, opened 2020-03-21
Ignore build directories that start with uppercase letters

php1ic:
This addresses #594

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .*
-build
-debug
-release
+[Bb]uild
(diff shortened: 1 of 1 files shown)


MaskRay:
This is correct.. but is it very necessary?

php1ic:
More of a personal preference than a necessity.

No problem if you don't want to do it, I wont be offended :wink:

MaskRay:
Let's just use a whitelist:)

closed by db12223081767fac2826512521bdeb61b6c01e88

Issue #555, opened 2020-01-04
Build failed at latest master branch


si1kdd:
Hi, i got a compiler error after these steps:

### command
```sh
git clone --depth=1 --recursive https://github.com/MaskRay/ccls && cd ccls
cmake -H. -BRelease -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH="$MY_LLVM_PATH"
```
## Output
```
-- The CXX compiler identification is GNU 9.1.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
-- Configuring done
-- Generating done
```

### command
```sh
cmake --build Release
```

## Output
```
Scanning dependencies of target ccls
[  2%] Building CXX object CMakeFiles/ccls.dir/third_party/siphash.cc.o
[  4%] Building CXX object CMakeFiles/ccls.dir/src/clang_tu.cc.o
[  6%] Building CXX object CMakeFiles/ccls.dir/src/config.cc.o
[  9%] Building CXX object CMakeFiles/ccls.dir/src/filesystem.cc.o
[ 11%] Building CXX object CMakeFiles/ccls.dir/src/fuzzy_match.cc.o
[ 13%] Building CXX object CMakeFiles/ccls.dir/src/main.cc.o
[ 16%] Building CXX object CMakeFiles/ccls.dir/src/include_complete.cc.o
[ 18%] Building CXX object CMakeFiles/ccls.dir/src/indexer.cc.o
/ccls/src/indexer.cc:684:8: error: ‘bool ccls::{anonymous}::IndexDataConsumer::handleDeclOccurrence(const clang::Decl*, clang::index::SymbolRoleSet, llvm::ArrayRef<clang::index::SymbolRelation>, clang::SourceLocation, clang::index::IndexDataConsumer::ASTNodeInfo)’ marked ‘override’, but does not override
[...]

PR #601, opened 2020-03-26
fix: compile error

ZhiruiLi:
handleDeclOccurrence(const clang::Decl*, clang::index::SymbolRoleSet, llvm::ArrayRef<clang::index::SymbolRelation>, clang::SourceLocation, clang::index::IndexDataConsumer::ASTNodeInfo)’ marked ‘override’, but does not override
   bool handleDeclOccurrence(const Decl *d, index::SymbolRoleSet roles,
        ^~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -685,7 +685,7 @@ public:
   bool handleDeclOccurrence(const Decl *d, index::SymbolRoleSet roles,
                             ArrayRef<index::SymbolRelation> relations,
                             SourceLocation src_loc,
-                            ASTNodeInfo ast_node) override {
+                            ASTNodeInfo ast_node) {
     if (!param.no_linkage) {
       if (auto *nd = dyn_cast<NamedDecl>(d); nd && nd->hasLinkage())
         ;


AraragiHokuto:
This compile error is caused by clang renaming the handleDeclOccurrence function (#555), causing a mismatch in name between parent and inherited class. I believe simply removing `override` does not really solve the problem?

ZhiruiLi:
> This compile error is caused by clang renaming the handleDeclOccurrence function (#555), which results in a mismatch in name between parent and inherited class. I believe simply removing `override` does not really solve the problem?

Actually it can solve the problem.

AraragiHokuto:
> > This compile error is caused by clang renaming the handleDeclOccurrence function (#555), which results in a mismatch in name between parent and inherited class. I believe simply removing `override` does not really solve the problem?
> 
> Actually it can solve the problem.

This seems to create an irrelevant new member function in the derived class, rather than overriding parent's. I'm not sure how removing `override` solve the problem (mismatch in member function name between parent class and derived class).

PR #617, opened 2020-04-21
workaround for building with clang 10

condy0919:


MaskRay:
Thanks for the pull request, but this is not correct. It is distributions' decisions to set `CLANG_LINK_CLANG_DYLIB` to `ON` or `OFF`. We can't assume every clang 10 package uses `-DCLANG_LINK_CLANG_DYLIB=on`.

If you know a clang>=10 package requires `-DCLANG_LINK_CLANG_DYLIB=on`, please help improve https://github.com/MaskRay/ccls/wiki/Build

PR #627, opened 2020-04-29
Minor modernization

condy0919:
Includes

- add `[[nodiscard]]` attribute for some functions
- change `jobs.size() == 0` to `jobs.empty()`
- remove `typename` in `void lock() { lock_impl(typename std::index_sequence_for<Queue...>{}); }`. There is no dependent type contextually.
- In `path = root + path.substr(real.size())` statement, `.substr()` returns a temporary `std::string`, and the `operator+` returns another one.  Let's optimize the second out.

### DIFF
--- a/src/utils.cc
+++ b/src/utils.cc
@@ -1,4 +1,4 @@
-// Copyright 2017-2018 ccls Authors
+// Copyright 2017-2020 ccls Authors
 // SPDX-License-Identifier: Apache-2.0
 
 #include "utils.hh"
@@ -37,7 +37,7 @@ Matcher::Matcher(const std::string &pattern)
                                         std::regex_constants::optimize);
 }
 
-Matcher::~Matcher() {}
+Matcher::~Matcher() = default;
 
 bool Matcher::matches(const std::string &text) const {
   return std::regex_search(text, impl->regex, std::regex_constants::match_any);
--- a/src/utils.hh
+++ b/src/utils.hh
@@ -1,4 +1,4 @@
-// Copyright 2017-2018 ccls Authors
+// Copyright 2017-2020 ccls Authors
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
@@ -25,7 +25,7 @@ struct Matcher {
   Matcher(const std::string &pattern); // throw
   Matcher(Matcher &&) = default;
   ~Matcher();
-  bool matches(const std::string &text) const;
+  [[nodiscard]] bool matches(const std::string &text) const;
 };
 
 struct GroupMatch {
@@ -114,7 +114,7 @@ public:
   const T &operator*() const { return storage; }
   T &operator*() { return storage; }
 
-  bool valid() const { return storage.valid(); }
+  [[nodiscard]] bool valid() const { return storage.valid(); }
   explicit operator bool() const { return valid(); }
   operator std::optional<T>() const {
     if (valid())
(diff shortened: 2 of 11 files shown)


src/utils.cc
```diff
@@ -140,7 +140,10 @@ std::string realPath(const std::string &path) {
 bool normalizeFolder(std::string &path) {
   for (auto &[root, real] : g_config->workspaceFolders)
     if (real.size() && llvm::StringRef(path).startswith(real)) {
-      path = root + path.substr(real.size());
+      // extra memory allocation may happen when string concatenates
```
OBorce, line 143:
```suggestion
      path.reserve(root.size() + path.size() - real.size());
```
why not reserve upfront since we already know the size?

src/utils.cc
```diff
@@ -140,7 +140,10 @@ std::string realPath(const std::string &path) {
...
+      path.clear();
+      path.append(root);
+      path.append(path.substr(real.size()));
```
OBorce, line 146:
```suggestion
      path.append(path, real.size());
```
no need for the substr temporary, append has an overload to append a sub-string directly

src/utils.cc
```diff
@@ -140,7 +140,7 @@ std::string realPath(const std::string &path) {
 bool normalizeFolder(std::string &path) {
   for (auto &[root, real] : g_config->workspaceFolders)
     if (real.size() && llvm::StringRef(path).startswith(real)) {
-      // extra memory allocation may happen when string concatenates
+      path.reserve(root.size() + path.size() - real.size());
       path.clear();
```
OBorce:
```suggestion
      path.clear();
      path.reserve(root.size() + path.size() - real.size());
```
on second look, it would be better to put the clear before the reserve, so that if reserve allocates it does not need to copy the old data in the new.

src/utils.cc
```diff
@@ -37,7 +37,7 @@ Matcher::Matcher(const std::string &pattern)
                                         std::regex_constants::optimize);
 }
 
-Matcher::~Matcher() {}
+Matcher::~Matcher() = default;
```
MaskRay, line 40: Not sure how `= default` helps.

src/utils.cc
```diff
@@ -140,7 +140,10 @@ std::string realPath(const std::string &path) {
 bool normalizeFolder(std::string &path) {
   for (auto &[root, real] : g_config->workspaceFolders)
     if (real.size() && llvm::StringRef(path).startswith(real)) {
-      path = root + path.substr(real.size());
```
MaskRay, line 143: This is hardly a bottleneck. The 4 lines are much harder to read.

src/utils.hh
```diff
@@ -25,7 +25,7 @@ struct Matcher {
   Matcher(const std::string &pattern); // throw
   Matcher(Matcher &&) = default;
   ~Matcher();
-  bool matches(const std::string &text) const;
+  [[nodiscard]] bool matches(const std::string &text) const;
```
MaskRay, line 28: How do you find places where `[[nodiscard]]` should be added? Did you just arbitrarily find some places and add `[[nodiscard]]`? (I am also less clear the attributes provides much value..)

condy0919: :rofl: They are diagnosed via `clang-tidy`

src/utils.cc
condy0919: I don't think it's hard to read.. A description comment may help.

src/utils.cc
condy0919: They are identical, using `= default` is more modern (diagnosed via `clang-tidy` with `modernize-use-default-member-init` option

FederAndInk: Can't we just let the compiler generate a destructor for us, and remove the declaration?

src/utils.cc
MaskRay: As discussed offline, "这就叫premature optimization" "找不出来为什么要扣这种地方" "就为了灌个contribution么……"

BTW, we can optimize your version further with `resize_default_init` http://wg21.link/p1072r5  ...

PR #314, opened 2019-03-10
If the workspace folder is a symlink, convert paths relative to it

MaskRay:
* The project root directory may be a symlink. The decision whether the symlink should be followed is on the client side.
  I believe many editors (emacs,vim) follow symlinks => there is no confusing part.

  If the editor doesn't (VSCode), ccls should use `/tmp/symlink/a.cc` as the canonical path, instead of `/tmp/real/a.cc`. In this case, the user had better not to open `/tmp/real/b.cc` directly (I'll call it "unsolicited open"), as t raises the question whether the symlink and the target should be considered the same file. The most sensible approach is probably to convert every `/tmp/real/` path to `/tmp/symlink/`.
* Some files/directories under the root directory are symlinks. If `/tmp/real/a.h -> /tmp/real/b.h` and `/tmp/real/a.cc` includes `a.h`. The clang internals always follow links and think the file is `b.h`. `a.h` and `b.h` have no distinction in clang. I'd like to treat `a.h` and `b.h` as different but it is impossible.

Issue #102, opened 2018-10-23
(Optionally) don't follow symlinks


sergei-dyshel:
I've noticed that `ccls` always resolves "real path" to referenced files by following symlinks. While there are certain advantages in this approach it may be undesirable in some cases. 

For instance, I'm using symlinks to disguise real location of some system header files which are located in different places on my working machines. When I transfer database from one machine to another, `ccs` have to recompile the whole database because of "missing" header files. 

I also believe a general argument: if files are referenced (in `compile_commands.json` or indirectly via include directives) with some symlink, there may be a reason for this and `ccls` should not be too smart and bypass this symlink by resolving the real location.

I think this behavior should be configurable. If author agrees, I may like to contribute a PR.

MaskRay:
Long ago I had the idea that symlinks under the project root should not be followed while others (e.g. `/usr/include/*` headers) should be followed.

I'd like to understand what kind of configuration you want to make first.

sergei-dyshel:
Ok, I'll try to be more specific.

I have a development server which has multiple versions of some library sitting in `/MyLib-1/`, `/MyLib-2/` etc. I want my project to compile against specific version but I don't want to hardcode library path in compilation flags so I make a symlink `/home/user/MyLib => /MyLib-2` and compile my project with `-I /home/user/MyLib/include`. 

Now I want to transfer a database indexed by `ccls` to my laptop. For this purpose I take the specific version of `MyLib` I need and fetch it to `/home/user/MyLib` on my laptop. The rest of file paths are same on the server and laptop so `ccls` index should be valid on laptop.

But it doesn't work currently, because when indexing on server `ccls` resolves each file in `/home/user/MyLib/include` to it's real location inside `/MyLib-2/include` so upon opening the index on laptop it can not find this location so it have to reindex everything.

So you might say why not store `/MyLib-1`, `/MyLib-2` on laptop too, but this might be not so easy if these libraries are part of some large directory structure which I don't need on laptop except for some header files needed by my project. So instead of partially recreating this directory structure I use symlinks which are universal UNIX solution for this kind of problems. 

Now after I told you my case can you please explain your reasoning of following symlinks for system headers?

MaskRay:
For a few system headers that can be references differently because of directory symlinks, tracing symlinks isn't bad, e.g. `/usr/include/c++/8/algorithm` `/usr/include/c++/8.0.1/algorithm`

That said, I agree with you that in most cases symlinks should not be followed.

[...]

PR #643, opened 2020-05-29
fix index error when there is a soft link file

Abbyyan:


MaskRay:
The expected behavior is to index the symlink. What particular problems do you want to address?

Abbyyan:
I have a directory whose structure is just like follows. 
image
When i use vscode to open `~/workspace/test1/test2.cpp`, its uri will be `/home/username/workspace/test1/test2.cpp`. But when i use ccls now, ccls will response `file not indexed` and all the lsp function can't worked. 

But now it seems the fix method has some problems. For example, when the uri in the request is a soft link `/home/username/workspace/test1/test2.cpp`, the uri in the response will be `/home/username/workspace/test2/test2.cpp`. The diagnostic will only appear on the file `/home/username/workspace/test2/test2.cpp`. Could you give me some suggestions about how to deal with softlink path files with ccls, please? Thanks a lot.

MaskRay:
See #102 and #314 for some discussions.

Issue #455, opened 2019-08-04
on mac, cmake module file for RapidJson provides RAPIDJSON_INCLUDE_DIRS instead of RapidJSON_INCLUDE_DIRS


Frefreak:
### Observed behavior

```
cmake .. -DUSE_SYSTEM_RAPIDJSON=ON
```
returns
```
-- RapidJSON found. Headers: /usr/local/Cellar/rapidjson/1.1.0/include
-- Configuring done
-- Generating done
```
but `make` returns error `fatal error: 'rapidjson/fwd.h' file not found`

The **CMakeLists.txt** assumes the include path of rapidjson be stored in `${RapidJSON_INCLUDE_DIRS}` but is actually `${RAPIDJSON_INCLUDE_DIRS}`. The relevant  cmake module is located in **/usr/local/lib/cmake/RapidJSON/RapidJSONConfig.cmake** and provided by rapidjson package (`brew install rapidjson`). Nothing big but definitely an issue which prevents the building from working "out of the box" on mac. 

### Expected behavior

build complete successfully.

### Steps to reproduce

```
cmake .. -DUSE_SYSTEM_RAPIDJSON=ON
make
```

### System information

* ccls version (`git describe --tags --long`): 0.20190314.1-21-g8b033217
* clang version: clang version 8.0.0 (tags/RELEASE_800/final)
* OS: MacOS mojave 10.14.6

MaskRay:
Duplicate of https://github.com/MaskRay/ccls/pull/383

PR #646, opened 2020-06-07
Fail loudly when found rapidjson is too old

vsui:
When the rapidjson found by cmake is an older version it defines a
variable `RAPIDJSON_INCLUDE_DIRS` instead of `RapidJSON_INCLUDE_DIRS` (#455).
According to #383 we do not want to make these older versions work with
ccls. However currently if the rapidjson found by cmake is an older
version that defined `RAPIDJSON_INCLUDE_DIRS` then the cmake invocation
still succeeds but the build command will fail because
`RapidJSON_INCLUDE_DIRS` was never set properly.

This PR makes the cmake invocation fail with a relevant error message in this event.

I think this would be useful because the homebrew version of rapidjson still defines RAPIDJSON_INCLUDE_DIRS so this issue is likely to come up for people on macOS. Especially since `-DUSE_SYSTEM_RAPIDJSON=on` is default now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,7 +152,7 @@ set_property(SOURCE src/utils.cc APPEND PROPERTY COMPILE_DEFINITIONS
 if(USE_SYSTEM_RAPIDJSON)
   find_package(RapidJSON QUIET)
   if(NOT DEFINED RapidJSON_INCLUDE_DIRS AND DEFINED RAPIDJSON_INCLUDE_DIRS)
-    message(FATAL_ERROR "RapidJSON version is likely too old. See https://github.com/MaskRay/ccls/issues/455.")
+    message(FATAL_ERROR "RapidJSON version is likely too old. See https://github.com/MaskRay/ccls/issues/455")
   endif()
 endif()
 if(NOT RapidJSON_FOUND)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,9 @@ set_property(SOURCE src/utils.cc APPEND PROPERTY COMPILE_DEFINITIONS
 
 if(USE_SYSTEM_RAPIDJSON)
   find_package(RapidJSON QUIET)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -151,6 +151,9 @@ set_property(SOURCE src/utils.cc APPEND PROPERTY COMPILE_DEFINITIONS
 
 if(USE_SYSTEM_RAPIDJSON)
   find_package(RapidJSON QUIET)
+  if(NOT DEFINED RapidJSON_INCLUDE_DIRS AND DEFINED RAPIDJSON_INCLUDE_DIRS)
+    message(FATAL_ERROR "RapidJSON version is likely too old. See https://github.com/MaskRay/ccls/issues/455.")
```
MaskRay, line 155: Drop the trailing period.

Xelbayria:
what's the solution? 

I used `cmake -H. -BRelease -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_PREFIX_PATH="C:/llvm/Release"
ninja -C Release`  and i got the same error where RapidJSON version is too old. Here's the error below: 
```
-- The CXX compiler identification is Clang 14.0.6 with MSVC-like command-line
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Scoop/apps/llvm/current/bin/clang-cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include pthread.h
-- Looking for C++ include pthread.h - not found
-- Found Threads: TRUE
-- Plase initialize rapidJSON git submodule as currently installed version is to old:
-- git submodule init && git submodule update
CMake Error at CMakeLists.txt:171 (message):
  RapidJSON version is likely too old.  See
  https://github.com/MaskRay/ccls/issues/455


-- Configuring incomplete, errors occurred!
See also "C:/ccls/Release/CMakeFiles/CMakeOutput.log".
See also "C:/ccls/Release/CMakeFiles/CMakeError.log".
```

PR #649, opened 2020-06-10
Build using LTO if compiler supports it (requires Cmake 3.9)

eklitzke:


MaskRay:
For clang, we have -flto=full and -flto=thin, and a bunch of other customizable options. Is it simpler delegating CXXFLAGS customization to the user? LTO may have some caveats.

Trass3r:
Does it make a difference given that ccls presumably links to the shared llvm lib?

PR #652, opened 2020-06-15
Do not use system RapidJSON by default

FederAndInk:
Now that we check some variables to see if rapidjson isn't too old and errors if this is the case #646, shouldn't we use the provided rapidjson by default? or at least not `message(FATAL_ERROR`?


xerrni:
#658 Fixes rapidJSON detection.

PR #656, opened 2020-06-28
 fix index error when there is a soft link file 

Abbyyan:
fix index error when there is a soft link file . Fix https://github.com/MaskRay/ccls/pull/643

### DIFF
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -25,6 +25,7 @@
 #include <inttypes.h>
 #include <map>
 #include <unordered_set>
+#include <filesystem>
 
 using namespace clang;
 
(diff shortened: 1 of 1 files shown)


MaskRay:
See #661

Abbyyan:
image

There are some error in the new version https://github.com/MaskRay/ccls/commit/99f0b402a7584f92de12f0af164ad7e0c772fcd2 . May be the change of worksapce cause it? It happens when i add `-v` and open a soft link file. `/home/Abbyyan/test/m1/bd -> /data/Abbyyan/test/m2/bd`. The error `file not found` will occur. 
```
"ccls.launch.args": [
        "-v=1",
    ],
```

PR #657, opened 2020-06-28
indexer: log the number of errors and the first diagnostic

MaskRay:
Example log:

```
15:47:45 indexer1         pipeline.cc:379 I parse /tmp/d/a.c error:1 use of undeclared identifier 'arg'
 clang /tmp/d/a.c --gcc-toolchain=/usr -working-directory=/tmp/d/
```

### DIFF
--- a/src/indexer.hh
+++ b/src/indexer.hh
@@ -322,13 +322,19 @@ struct IndexFile {
   std::string toString();
 };
 
+struct IndexResult {
+  std::vector<std::unique_ptr<IndexFile>> indexes;
+  int n_errs = 0;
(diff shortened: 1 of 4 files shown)


src/indexer.hh
```diff
@@ -322,13 +322,19 @@ struct IndexFile {
   std::string toString();
 };
 
+struct IndexResult {
+  std::vector<std::unique_ptr<IndexFile>> indexes;
+  int n_errs;
```
hotpxl, line 327: specify default value =0, especially since you have `return {}` in the above function

PR #660, opened 2020-07-02
fix outline and breadcumbs for members definition in namespace

FederAndInk:
When we have members defined outside the class in a namespace in a separate file, the namespace doesn't have the members as children. The outline and breadcrumbs don't work correctly for them.
Before:
Screenshot_20200702_214411
After:
Screenshot_20200702_213950

I also refactored a little by removing `void*` and replacing it with `if constexpr` see the first commit if it suits your requirements.

There are other bugs with outline that I noticed which I am working on:
- [x] static member variable aren't in their class/struct
- [ ] defined member functions after the class definition in the same file override the member function declaration
- [ ] inline namespace

### DIFF
--- a/src/indexer.hh
+++ b/src/indexer.hh
@@ -65,6 +65,8 @@ struct SymbolIdx {
   }
 };
 
+REFLECT_STRUCT(SymbolIdx, usr, kind);
+
 // |id,kind| refer to the referenced entity.
 struct SymbolRef {
   Range range;
@@ -163,6 +165,7 @@ struct FuncDef : NameMixin<FuncDef<V>> {
   int16_t short_name_size = 0;
   SymbolKind kind = SymbolKind::Unknown;
   SymbolKind parent_kind = SymbolKind::Unknown;
+  SymbolIdx parent = {0, Kind::Invalid};
   uint8_t storage = clang::SC_None;
 
   const Usr *bases_begin() const { return bases.begin(); }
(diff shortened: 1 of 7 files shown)


FederAndInk:
Ok, static member variable aren't collected in `collectRecordMembers` because `rd->fields()` gives non-static data members only. Thus there aren't put in `type.def.vars`. at [src/indexer.cc:673](https://github.com/MaskRay/ccls/blob/468258d641391015318bdd6ef0c8d06543acf6ee/src/indexer.cc#L673)

Also, another bug I just discovered, I haven't looked up in details:
- inline namespace

FederAndInk:
static member variable fixed:
Before:
Screenshot_20200704_130027
After:
Screenshot_20200704_130045

src/indexer.hh
```diff
@@ -163,14 +165,15 @@ struct FuncDef : NameMixin<FuncDef<V>> {
   int16_t short_name_size = 0;
   SymbolKind kind = SymbolKind::Unknown;
   SymbolKind parent_kind = SymbolKind::Unknown;
+  SymbolIdx parent = {0, Kind::Invalid};
```
MaskRay, line 168: Will this increase memory usage?

FederAndInk: Yeah, certainly, I haven't done measures. But I am working on another idea which doesn't introduce new things to the index and seems simpler and cleaner.

MaskRay:
Thanks for this patch and the alternative:) Merged the alternative

PR #661, opened 2020-07-04
Index symlinks by replacing tryGetRealName with getName+normalization

### LINKED ISSUE #639: Error when there is soft link in the project
When there are soft link in my project ， the message return will become 
`"error":{"code":-32600,"message":"not indexed"}`

MaskRay:
Fix #639


MaskRay:
Merged.

PR #666, opened 2020-07-08
fix when workspace has soft link; eg: /home/Abbyyan->/data/home/Abbya…

Abbyyan:
Fix with 99f0b402a7584f92de12f0af164ad7e0c772fcd2 。
I test on 99f0b402a7584f92de12f0af164ad7e0c772fcd2 and it may fail in such situation. 
Workspace `/home/Abbyyan/workspace` is a soft link and links to `/data/Abbyyan/workspace`， while the opened file `/home/Abbyyan/workspace/test1/main.cpp ` also  has soft link  whose dir `test1` links to `test2`. 
```shell
/home/Abbyyan/workspace -> /data/Abbyyan/workspace
/home/Abbyyan/workspace/test1/main.cpp -> /data/Abbyyan/workspace/test2/main.cpp
```
In this case, the function `normalizeFolder` in `src/utils.cc`  https://github.com/MaskRay/ccls/blob/99f0b402a7584f92de12f0af164ad7e0c772fcd2/src/utils.cc#L140 should return `True` to use the soft link path . While now it will return False and get the absolute path. So when i open with `/home/Abbyyan/workspace/test1/main.cpp ` , the index can't be find.  Maybe it should change as follows.
```c++
bool normalizeFolder(std::string &path) {
  for (auto &[root, real] : g_config->workspaceFolders) {
    if (StringRef(path).startswith(real)) {
      path = root + path.substr(real.size());
      return true;
    } else if (StringRef(path).startswith(root)) {
      return true;
    }
  }

  return false;
}
```
This function is also used in other places.  I am not sure if it will affect other places, so i add a new function in this pull request.

### DIFF
--- a/src/clang_tu.cc
+++ b/src/clang_tu.cc
@@ -17,13 +17,13 @@
 #include <llvm/Support/Path.h>
 
 using namespace clang;
+using namespace llvm;
 
 namespace ccls {
 std::string pathFromFileEntry(const FileEntry &file) {
-  StringRef name = file.tryGetRealPathName();
-  if (name.empty())
-    name = file.getName();
-  std::string ret = normalizePath(name);
+  SmallString<128> path(file.getName());
+  sys::path::remove_dots(path, /*remove_dot_dot=*/true);
+  std::string ret(path.str());
   // Resolve symlinks outside of workspace folders, e.g. /usr/include/c++/7.3.0
   return normalizeFolder(ret) ? ret : realPath(ret);
 }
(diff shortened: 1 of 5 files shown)


MaskRay:
Thanks. Should be fixed by 90a596cb0cc3d84a3bd0ad304c3e57f535251622. Sorry about the forced push.. (This kind of stuff is tricky and I don't want many commits to cancel each other and clutter up the history)

PR #673, opened 2020-07-20
clang format, indent pp directives

FederAndInk:
I saw this style for pp directives in the code:
```cpp
#if LLVM_VERSION_MAJOR < 10 // llvmorg-10-init-12036-g3b9715cb219
#  define handleDeclOccurrence handleDeclOccurence
#endif
```

and there is an option for it in clang-format if you want, it also touches includes which looks weird?
I also ran clang-format on all files.

Or we can go with the default style but we have to run clang-format anyway because some files aren't formatted correctly (indexer.cc, clang_tu.cc).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,1 +1,2 @@
 BasedOnStyle: LLVM
+IndentPPDirectives: AfterHash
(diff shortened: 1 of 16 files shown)


MaskRay:
Thanks for the patch. I recently learned `IndentPPDirectives: AfterHash` as well but it seems a bit of a churn to format all existing files.

PR #674, opened 2020-07-20
Fix document symbol with a range based method (outline and breadcumbs)

FederAndInk:
**This is an alternative to https://github.com/MaskRay/ccls/pull/660, I have redone all the algorithm for hierarchical document symbol, it is based on ranges of symbols and not on symbols and their children because some of them were missing and "duplicates" were removed (eg: namespace redefinition in the same file, or inline function defined after the class)** 

When we have members defined outside the class in a namespace in a separate file, the namespace doesn't have the members as children. The outline and breadcrumbs don't work correctly for them.
Before:
Screenshot_20200702_214411
After:
Screenshot_20200702_213950

There are other bugs with outline that I noticed which I am working on:
- [x] static member variable aren't in their class/struct
- [x] defined member functions after the class definition in the same file override the member function declaration
- [x] inline namespace

### DIFF
--- a/src/messages/textDocument_document.cc
+++ b/src/messages/textDocument_document.cc
@@ -156,32 +156,25 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
     syms.reserve(file->symbol2refcnt.size());
 
     /**
-     * std::heap is a max heap (we swap lhs/rhs to make a min heap)
-     *
      * with 2 ranges that start at the same Position, we want the wider one
-     * first
-     *
+     * first (swap lhs/rhs)
      */
     auto sym_cmp = [](ExtentRef const &lhs, ExtentRef const &rhs) {
-      return rhs.extent.start < lhs.extent.start ||
-             (rhs.extent.start == lhs.extent.start &&
-              lhs.extent.end < rhs.extent.end);
+      return lhs.extent.start < rhs.extent.start ||
+             (lhs.extent.start == rhs.extent.start &&
+              rhs.extent.end < lhs.extent.end);
     };
 
     for (auto [sym, refcnt] : file->symbol2refcnt) {
       if (refcnt <= 0 || !sym.extent.valid())
         continue;
       syms.push_back(sym);
-      std::push_heap(std::begin(syms), std::end(syms), sym_cmp);
     }
+    std::sort(std::begin(syms), std::end(syms), sym_cmp);
 
     std::vector<DocumentSymbol> result;
     std::stack<DocumentSymbol *> indent;
-    while (!syms.empty()) {
-      std::pop_heap(std::begin(syms), std::end(syms), sym_cmp);
-      auto sym = syms.back();
-      syms.pop_back();
-
+    for (auto sym : syms) {
       DocumentSymbol ds;
       if (auto range = getLsRange(wf, sym.range)) {
         ds.selectionRange = *range;
@@ -194,7 +187,7 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
               range1 && range1->includes(*range))
             ds.range = *range1;
       }
-      withEntity(db, sym, [&, sym = sym](const auto &entity) {
+      withEntity(db, sym, [&](const auto &entity) {
         auto const *def = entity.anyDef();
         if (!def)
           return;
--- a/src/messages/textDocument_document.cc
+++ b/src/messages/textDocument_document.cc
@@ -6,6 +6,7 @@
 #include "query.hh"
 
 #include <algorithm>
+#include <stack>
 
 MAKE_HASHABLE(ccls::SymbolIdx, t.usr, t.kind);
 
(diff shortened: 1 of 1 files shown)


src/messages/textDocument_document.cc
```diff
@@ -165,88 +152,64 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
     std::sort(result.begin(), result.end());
     reply(result);
   } else if (g_config->client.hierarchicalDocumentSymbolSupport) {
-    std::unordered_map<SymbolIdx, std::unique_ptr<DocumentSymbol>> sym2ds;
-    std::vector<std::pair<std::vector<const void *>, DocumentSymbol *>> funcs,
-        types;
+    std::vector<ExtentRef> syms;
```
FederAndInk, line 155: We may want this to be a vector of pointer to take less memory 32B to 8B per element and then sort the pointers.

src/messages/textDocument_document.cc
```diff
@@ -165,88 +152,64 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
...
+    syms.reserve(file->symbol2refcnt.size());
```
FederAndInk, line 156: Here I reserve probably too much space but at least there is only 1 allocation done. What do you think?

FederAndInk:
Is there someone who can review this?

joelhock:
i can say this fixed the issue for me, so I'm in favor!

src/messages/textDocument_document.cc
```diff
@@ -165,88 +152,70 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
...
+                 // issue with: int i, j, k; otherwise
+                 indent.top()->range.end < ds.selectionRange.start) {
+            indent.pop();
           }
+          if (indent.empty()) {
+            result.push_back(std::move(ds));
+            indent.push(&result.back());
+          } else {
+            indent.top()->children.push_back(std::move(ds));
+            indent.push(&indent.top()->children.back());
```
MaskRay, line 214: There is a pointer invalidation issue. I added `    result.reserve(syms.size());`

src/messages/textDocument_document.cc
```diff
@@ -165,88 +152,70 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
...
-            kind = ds->kind = def.kind;
-            def_ptrs.push_back(&def);
+        ds.name = def->name(false);
+        ds.detail = def->detailed_name;
+        ds.kind = def->kind;
+
+        if (!ignore(def) && (ds.kind == SymbolKind::Namespace || allows(sym))) {
+          // drop symbols that are behind the current one
+          while (!indent.empty() &&
+                 // use selectionRange for start to start at symbol name
```
MaskRay, line 204: Elaborate this comment a bit? How does `int i, j, k;` cause an issue?

FederAndInk: When we define multiple variable on "the same line" eg: int i, j, k;
k is child of j and j is child of i (if we use `range`)

This is because `range` starts at the type so at the same position for i, j and k, but with `selectionRange` it starts at the name of the variable thus we can separate the 3 and not make j a child of i.

The outline before:
k
 j
  i
after:
i
j
k

src/messages/textDocument_document.cc
MaskRay: I switched back to `unique_ptr`

FederAndInk: Oh yeah, thanks, such a common issue, that's why I should have used a vector of pointers.

src/messages/textDocument_document.cc
MaskRay: I don't understand. With the current `while (!scopes.empty() && scopes.back()->range.end <= ds->selectionRange.start)`, i is the parent of j and j is the parent of k. Only by switching to `ds->range.start` we can make them flattened.

src/messages/textDocument_document.cc
```diff
@@ -165,88 +152,70 @@ void MessageHandler::textDocument_documentSymbol(JsonReader &reader,
...
+
+    /**
+     * with 2 ranges that start at the same Position, we want the wider one
+     * first (swap lhs/rhs)
+     *
+     * we use range for start to start at symbol name
+     * issue with: int i, j, k; otherwise
+     */
+    auto sym_cmp = [](ExtentRef const &lhs, ExtentRef const &rhs) {
+      return lhs.range.start < rhs.range.start ||
```
MaskRay, line 166: I am currently using
```cpp
                return std::tie(lhs.range.start, rhs.extent.end) <
                       std::tie(rhs.range.start, lhs.extent.end);
```

but why not
```cpp
                return std::tie(lhs.extent.start, rhs.extent.end) <
                       std::tie(rhs.extent.start, lhs.extent.end);
```

src/messages/textDocument_document.cc
FederAndInk: `ds->range` is `sym.extent` if valid (it's a little bit confusing but that's how it was) and sym.extent begins at the type so here: `int i, j, k;`
with `ds->selectionRange` (which is `sym.range`) it begins at i or j or k

I don't fully understand where we disagree

src/messages/textDocument_document.cc
FederAndInk: This is the same problem as under: https://github.com/MaskRay/ccls/pull/674/commits/48b51507fdfeda0d4a23014b08e2683d8ddc72d0#r628918008
`extent.start` points to the type, and `range.start` points to the variable name

to put i then j then k flattened we need different start position

FederAndInk: Sorry, it's there to sort `int i, j, k` in the correct order, or else it keeps the order of `file->symbol2refcnt` as `extent.start` of i, j and k is the same value (points to `int`) but `range.start` points to the variable name.

MaskRay: OK, you meant global variables `int i, j, k;` I understand that now. You are right, sorting by `range.start` can fix the nested k,j,i problem.

Fixed. Thanks!

src/messages/textDocument_document.cc
```diff
@@ -6,6 +6,7 @@
 #include "query.hh"
 
 #include <algorithm>
+#include <stack>
```
MaskRay, line 9: I removed this as well. We can just use `vector`.

src/messages/textDocument_document.cc
MaskRay: This is fine. syms is about the same scale of symbol2refcnt.

src/messages/textDocument_document.cc
FederAndInk: good idea

src/messages/textDocument_document.cc
FederAndInk: or static/member variables

PR #700, opened 2020-09-23
Fix a trivial constructor argument change from upsream

lanza:


lanza:
https://reviews.llvm.org/rG257b29715bb

MaskRay:
Thanks. https://reviews.llvm.org/D86089

We cannot simply change the signature because we need to be compatible with some older release (7, 8, 9, 10; 7 is still there because maintaining it is not too burdensome). I will prepare a change soon.

MaskRay:
1dc69f45dc3e0c0f750f5a2e220a6aea54f30d2a

Issue #703, opened 2020-10-06
Relative "directory" in compile_commands.json does not work if CWD != project root


rohit04saluja:
**Observed behavior**
ccls does not work when directory value in compile_commands.json is relative to the root (workspace). ccls works with directory value in compile_commands.json as absolute.

**Expected behavior**
If directory value in compile_commands.json is a relative to root (workspace) path, then it should be resolved using the root.

**Steps to reproduce**
Just give directory a relative value w.r.t root in compile_commands.json.

System information
ccls version: 0.1.28
OS: Linux
Editor: vs-code

I was able to see in the code that when we are getting the directory information we don't check for the relative path as we do for _filename_.

// If workspace folder is real/ but entries use symlink/, convert to // real/. entry.directory = realPath(cmd.Directory); normalizeFolder(entry.directory); doPathMapping(entry.directory); entry.filename = realPath(resolveIfRelative(entry.directory, cmd.Filename)); normalizeFolder(entry.filename); doPathMapping(entry.filename);

Just like we resolve value if it is relative, for filename with the anchor as directory, we should resolve if relative for directory as well with the anchor as root.

MaskRay:
```json
% cat compile_commands.json 
[
    {
        "arguments": [
            "g++",
            "-DA",
            "-c",
            "-o",
            "a",
            "a.cc"
        ],
        "directory": ".",
        "file": "a.cc"
    }
]
```

This works fine.

rohit04saluja:
Hey @MaskRay I am using a similar compile_commands.json with the "command" key in place of the "arguments" key in vscode, but it's not working. Any leads on how I can debug it?

MaskRay:
`"command"` should work as well. https://github.com/MaskRay/ccls/wiki/Debugging has some tips for debugging. In particular, you can set up a breakpoint on `Project::findEntry` before opening a file.

rohit04saluja:
The issue is that VSCode does not chdir to the project root. Thus directory path relative to users directory works fine as  mentioned above

MaskRay:
Thanks for confirmation. #711

rohit04saluja:
@MaskRay should we fix the change directory bug in the VSCode client as well?

PR #704, opened 2020-10-06
Allow relative path for directory

rohit04saluja:
directory in _compiler_commands.json_
If the value is relative we resolve it from the root

This will help in VS Code integration. We open the root folder in **VS Code** and the directory value can be specified as related to the root folder. Thus the _compile_commnads.json_ can be committed to the code repository rather than having to build it every time code is pulled.

### DIFF
--- a/src/clang_tu.cc
+++ b/src/clang_tu.cc
@@ -111,7 +111,11 @@ buildCompilerInvocation(const std::string &main, std::vector<const char *> args,
   IntrusiveRefCntPtr<DiagnosticsEngine> diags(
       CompilerInstance::createDiagnostics(new DiagnosticOptions,
                                           new IgnoringDiagConsumer, true));
+#if LLVM_VERSION_MAJOR < 12 // llvmorg-12-init-5498-g257b29715bb
   driver::Driver d(args[0], llvm::sys::getDefaultTargetTriple(), *diags, vfs);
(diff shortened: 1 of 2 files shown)


rohit04saluja:
**Linked issue**: directory value as relative path in compile_commands.json #703 [issue](https://github.com/MaskRay/ccls/issues/703)

MaskRay:
Currently `realPath` resolves the "directory" value relative to the current working directory. Is current working directory unchanged in VSCode?

rohit04saluja:
Assuming that the current working directory refers to the workspace directory. I tried all possible combinations but it didn't work.

In CCLS logs, I can see that the files are resolved with relative paths only.
With this fix, it's resolved properly

rohit04saluja:
Hi @MaskRay did you get time to check this?

MaskRay:
I believe the problem is that your client does not chdir to the project root. Which client do you use?

rohit04saluja:
I use VSCode

rohit04saluja:
You were right @MaskRay . VSCode does not chdir to the project root It's always pointing to my home directory.

But since ccls is given a project path to initialize, won't be it good to chdir to the project path during initialization?

rohit04saluja:
Need to be looked into from another angle

PR #711, opened 2020-10-27
Resolve "directory" in compile_commands.json in case CWD is not the project root

MaskRay:
Also delete an unneeded chdir hack.


MaskRay:
ccls supports workspace folders. With workspace folders, you may have several projects. The snippet does not work anyway.

rohit04saluja:
Do you mind if I try doing this enhancement/bug fix?

PR #721, opened 2020-11-13
Add -m64 to gcc by default on illumos

AraragiHokuto:
GCC on illumos generates 32-bit object files by default to maintain
backward compatibility, despite 32-bit platforms are no longer
supported. This may cause link errors, since most distros seem to supply
64-bit LLVM in their repositories (verified on OpenIndiana and OmniOS).
This commit adds an ILLUMOS_USE_M64 option, which is turned on by
default, to instruct gcc on illumos to generate 64-bit object files, as
this is usually the desired behaviour.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,23 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL FreeBSD)
   target_link_libraries(ccls PRIVATE thr)
 endif()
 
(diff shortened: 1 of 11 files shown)


AraragiHokuto:
Note: This change added a illumos-specific option (ILLUMOS_USE_M64), which probably should be documented in wiki if this PR gets merged.

PR #740, opened 2020-12-25
First attempt to include the semantic highlight feature, introduced in specification 3.16.0.

jlahoda:
This is a crude attempt to include semantic highlights introduced in LSP spec 3.16.0:
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens

It is mostly copied src/message_handler.cc#emitSemanticHighlight, with some tweaks. Feedback is welcome!

### DIFF
--- a/src/messages/textDocument_semanticToken.cc
+++ b/src/messages/textDocument_semanticToken.cc
@@ -1,0 +1,259 @@
+// Copyright 2017-2020 ccls Authors
+// SPDX-License-Identifier: Apache-2.0
+
+#include "indexer.hh"
+#include "message_handler.hh"
+#include "pipeline.hh"
+#include "sema_manager.hh"
+
+#include <clang/Sema/Sema.h>
+#include <algorithm>
+#include <stdexcept>
+#include <string>
+
+#include <iostream>
+
+MAKE_HASHABLE(ccls::SymbolIdx, t.usr, t.kind);
+
+namespace ccls {
+using namespace clang;
+
+namespace {
+struct SemanticTokens {
+  std::vector<int> data;
+};
+REFLECT_STRUCT(SemanticTokens, data);
+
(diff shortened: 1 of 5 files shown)


MaskRay:
Thanks! I guess the specification is probably a bit unfortunate, since the payload may be much larger than the ccls extension...

FelipeLema:
I've been using this branch today and it looks like it works all right (at least, for starters… I didn't see any obvious errors or problems)

@MaskRay is there any documentation of how to handle the missing `TODO`s? (I didn't seem to run to any of these errors, btw)

src/messages/textDocument_semanticToken.cc
```diff
@@ -0,0 +1,257 @@
...
+      return symbol->kind < o.symbol->kind;
+    // If symbol A and B occupy the same place, we want one to be placed
+    // before the other consistantly.
+    return symbol->id < o.symbol->id;
+  }
+};
+
+}
+
+void MessageHandler::textDocument_semanticTokensFull(
```
MaskRay, line 57: This replicates too many lines in `message_handler.cc`. You can make the `message_handler.cc` implementation public and call that function.

MaskRay (changes requested):
Do you know which client can be used for testing? It seems that semantic tokens do not support rainbow semantic highlighting?

src/messages/textDocument_semanticToken.cc
FelipeLema: I couldn't find a quote, but as far as I understand `emitSemanticHighlight()` is being replaced by Semantic Tokens.

I remember reading about it in clangd issues, but I cannot find a hard quote on this. Best mention is done [here](https://github.com/clangd/clangd/issues/352#issuecomment-619442147)

I'll remove old semantic highlighting code.

PR #745, opened 2020-12-30
Initialization Option: completion.overwriteFollowingIdentifier

### LINKED ISSUE #678: Completion overwriting word
### Observed behavior
I'm not quite sure where this is configured or what it stems from, but if I switch to ```clangd``` the behavior is as expected. Also not sure if this is something people generally want and I'm weird for reporting it, but here goes:
When selecting a completion, the selected completion overwrites the current word my cursor is on.
Screenshot_20200725_100912
Screenshot_20200725_100935
Screenshot_20200725_100951

### Expected behavior

Insert the completion candidate, rather than overwriting the word at the cursor.

### Steps to reproduce

1. Go to any word
2. Type and select a completion option

### System information
ccls version 0.20190823.6-22-gcb08df44
clang version 10.0.0 
* OS: Arch Linux (Linux Office-Arch 5.7.9-arch1-1 #1 SMP PREEMPT Thu, 16 Jul 2020 19:34:49 +0000 x86_64 GNU/Linux)
* Editor: Emacs
* Language client (and version): lsp-mode (Not sure where the version is, I use straight.el, so here's the last commit in the repo)
```
lsp-mode: git log
commit 41ba941797f5c0cb9b251af5ea37922595d772bc (HEAD -> master, origin/master, origin/HEAD)
Author: kiennq <kien.n.quang@gmail.com>
Date:   Fri Jul 24 18:42:10 2020 +0900

    type-formatting: do best effort async text edits applying (#1959)
```

AraragiHokuto:
Previously, ccls overwrites the identifier immediately following the
cursor on completions. Some user does not want this behaviour.

This commit introduces a new initialization option:
completion.overwriteFollowingIdentifier, which allows
customizing the said behaviour.

fix #678.

### DIFF
--- a/src/working_files.cc
+++ b/src/working_files.cc
@@ -347,9 +347,12 @@ Position WorkingFile::getCompletionPosition(Position pos, std::string *filter,
     --i;
 
   *replace_end_pos = pos;
-  for (int i = start;
-       i < buffer_content.size() && isIdentifierBody(buffer_content[i]); i++)
-    replace_end_pos->character++;
+  if (g_config->completion.overwriteFollowingIdentifier) {
+    // attempt to overwrite the following identifier
+    for (int i = start;
+         i < buffer_content.size() && isIdentifierBody(buffer_content[i]); i++)
+      replace_end_pos->character++;
+  }
 
   *filter = buffer_content.substr(i, start - i);
   return getPositionForOffset(buffer_content, i);
--- a/src/config.hh
+++ b/src/config.hh
@@ -166,6 +166,11 @@ struct Config {
     // that implement their own filtering and sorting logic.
     bool filterAndSort = true;
 
+    // Decides whether to overwrite the identifier following cursor.
+    // true -- overwrite the following identifier.
+    // false -- insert completion before the following identifier.
+    bool overwriteFollowingIdentifier = true;
+
     struct Include {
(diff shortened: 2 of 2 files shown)


src/working_files.cc
```diff
@@ -346,10 +346,15 @@ Position WorkingFile::getCompletionPosition(Position pos, std::string *filter,
...
-       i < buffer_content.size() && isIdentifierBody(buffer_content[i]); i++)
-    replace_end_pos->character++;
+  if (g_config->completion.overwriteFollowingIdentifier) {
+    *replace_end_pos = pos;
+    for (int i = start;
+         i < buffer_content.size() && isIdentifierBody(buffer_content[i]); i++)
+      replace_end_pos->character++;
+  } else {
+    // do not overwrite the following identifier
+    *replace_end_pos = pos;
```
madscientist, line 356: This could be pulled out before the if-statement, then you don't need the else block at all.

AraragiHokuto: Done. Thanks for spotting that.

MaskRay:
Thanks for the patch. I also asked some other users and it seems that most people think the following identifier should not be overwritten. I think not overwriting the identifier is an option the client can provide instead. So #749 just switched the default behavior.

Issue #720, opened 2020-11-12
Build failure on windows with llvm 11.0


bin:
I am attempting to build ccls on windows 10 with the latest release 0.20201025, Clang and LLVM 11.0, and Visual Studio 2019.  When compiling, I receive the following error:

```
ccls\src\messages\workspace.cc(41,79): error C2679: binary '=': no operator found which takes a right-hand operand of type 'llvm::StringRef' (or there is no acceptable
conversion)
```

Let me know if I can send any more information to help.

jamiltron:
I am also receiving this error, unfortunately. Similar setup.

MaskRay:
```bash
cmake -GNinja -H. -BBundled11 -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=$HOME/llvm-prebuilt/clang+llvm-11.0.0-x86_64-linux-gnu-ubuntu-20.04
ninja -C Bundled11
```

It works on Linux. I suspect `StringRef` does not implicitly convert to `std::string_view` on MSVC.

anurag-23:
Please refer to https://github.com/MaskRay/ccls/pull/747#issuecomment-762045660

This is an issue with MSVC incorrectly reporting the value of __cplusplus

MaskRay:
Fixed by #747 . Thanks everyone participating in the discussion:)

PR #747, opened 2021-01-07
Make llvm::StringRef to std::string conversions explicit.

l1nxy:
The implicit convertion can not be compiled in win10, llvm 11.0, msvc 2019.
Same issue there: #720

### DIFF
--- a/src/clang_tu.cc
+++ b/src/clang_tu.cc
@@ -25,7 +25,7 @@ std::string pathFromFileEntry(const FileEntry &file) {
     // Resolve symlinks outside of working folders. This handles leading path
     // components, e.g. (/lib -> /usr/lib) in
     // /../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/utility
-    ret = file.tryGetRealPathName();
+    ret = std::string(file.tryGetRealPathName());
   } else {
     // If getName() refers to a file within a workspace folder, we prefer it
     // (which may be a symlink).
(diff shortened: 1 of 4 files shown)


MaskRay:
```cpp
#include <string>
#include <string_view>

struct StringRef {
    const char *s;
    size_t len;
    operator std::string_view() const {
        return {s, len};
    }
};

void foo() {
    std::string s;
    s = StringRef{"a", 1};
}
```

This compiles with `/std:c++17`. Could you check what happens?

anurag-23:
I was running into the same issue on Visual Studio 2019.

For me, the directive `#if __cplusplus > 201402L` was strangely evaluating to **_false_** in _LLVM\include\llvm\ADT\StringRef.h_

On checking the value,  __cplusplus is being defined as **_199711L_**, despite using /std:c++17

Because of this the following were not being compiled:

` #if __cplusplus > 201402L
 #include <string_view>
 #endif`

` #if __cplusplus > 201402L
     /// Construct a string ref from an std::string_view.
     /*implicit*/ constexpr StringRef(std::string_view Str)
         : Data(Str.data()), Length(Str.size()) {}
 #endif
 `

` #if __cplusplus > 201402L
     operator std::string_view() const {
       return std::string_view(data(), size());
     }
 #endif`

Once I removed these directives, ensuring that these fragments of code are compiled, the build was successful. Seems like an issue with MSVC where the __cplusplus value is being incorrectly reported.

Issue #768, opened 2021-02-11
Generated json contains \u0000 and causes lsp-mode / emacs-ccls to fail parsing


triplejam:
### Observed behavior

When lsp-treemacs-symbols is called, or if lsp-ui is installed in emacs, certain c++ files will cause emacs to show a warning about failing to parse json. If lsp-treemacs-symbols was called it will be empty. 

I think it is related to user-defined conversion functions and or typedefs.

### Expected behavior

JSON would be parsed successfully and there would be no warning. The symbols tree would be complete.

### Steps to reproduce

Small example which reproduces the issue:

```cpp 
struct A
{
    typedef unsigned int sz;
    operator sz() const { return 0; }
}; 
```

If `int` is used instead of `unsigned int` it works fine. Alternatively if `explicit` is put before operator it works.

### System information

* ccls version 0.20201219-5-g4cd308d7
* clang version: 11.0.1
* OS: OSX 11.2.1
* Editor: Emacs 27.1
* Language client (and version): ccls 20200820.308

TheJJ:
I got the same on latest git 4cd308d7.

It looks like some out-of-bounds string read.

```cpp
class ... {
[snip]
	operator T() const {
		return this->value;
	}
[snip]
}
```

```python
 {'children': [],
  'detail': 'operator T() const',
  'kind': 6,
  'name': 'operator T() const\x00std::__s',
  'range': {'end': {'character': 2, 'line': 63},
            'start': {'character': 1, 'line': 61}},
  'selectionRange': {'end': {'character': 11, 'line': 61},
                     'start': {'character': 1, 'line': 61}}},
```

the json contains the 0-byte, likely it's the terminator of the function name, since there's more stuff starting after it.

orlandini:
While this is not solved, the following snippet in , originally posted by @yyoncho in https://github.com/emacs-lsp/lsp-mode/issues/2681#issuecomment-792563858, is a workaround:
```
(advice-add 'json-parse-string :around 
            (lambda (orig string &rest rest)
              (apply orig (s-replace "\\u0000" "" string)
                     rest)))
```

madscientist:
It seems like that snippet isn't really quite correct: it surely will get rid of the warnings but if the problem is that extra text is being included the more correct thing would be to remove all text starting with \00 up to the end of the string, rather than deleting the \00 by itself.  Otherwise the "name" field will contain garbage.

[...]

PR #784, opened 2021-03-30
Fix short_name_size when getNameAsString does not return a prefix

zhangjunphy:
This is an attempt to fix https://github.com/MaskRay/ccls/issues/768.
Sets `short_name_size` to the length of the full identifier name when clang function `getNameAsString` does not return a prefix of it, preventing later out-of-bound reads in `NameMixin::name`.

### DIFF
--- a/src/indexer.cc
+++ b/src/indexer.cc
@@ -534,15 +534,18 @@ public:
       // e.g. operator type-parameter-1
       i = 0;
       def.short_name_offset = 0;
-    } else if (short_name.empty() || (i >= 2 && name[i - 2] == ':')) {
-      // Don't replace name with qualified name in ns::name Cls::*name
-      def.short_name_offset = i;
+      def.short_name_size = name.size();
     } else {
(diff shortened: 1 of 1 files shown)


src/indexer.cc
```diff
@@ -687,7 +690,7 @@ class IndexDataConsumer : public index::IndexDataConsumer {
   IndexDataConsumer(IndexParam &param) : param(param) {}
   void initialize(ASTContext &ctx) override { this->ctx = param.ctx = &ctx; }
 #if LLVM_VERSION_MAJOR < 10 // llvmorg-10-init-12036-g3b9715cb219
-# define handleDeclOccurrence handleDeclOccurence
+#define handleDeclOccurrence handleDeclOccurence
```
MaskRay, line 693: Drop this difference...

zhangjunphy: Done. Must be sneaked in by the formatter, didn't notice it. 😅

MaskRay:
Thanks!

PR #794, opened 2021-05-09
remove document symbols from macro expansion

FederAndInk:
There is one issue in #674 though that I've completely forgotten about, it does not manage macro expansion eg:
```cpp
#define M                                                                                \
  namespace ns_name                                                                      \
  {                                                                                      \
  namespace ns_name_impl                                                                 \
  {                                                                                      \
  void f() {}                                                                            \
  }                                                                                      \
  }

M;
```

may generate this outline:
image

clangd is showing symbols in macro definition but not in macro expansion

at least we have to exclude symbols from macro expansion, here is a quick PR to detect and remove macro expansion, but it may exist a better implementation

The question is, what to do with macro expansion? Should we remove them? (like this PR does)

### DIFF
--- a/src/messages/textDocument_document.cc
+++ b/src/messages/textDocument_document.cc
@@ -130,6 +130,20 @@ template <> bool ignore(const QueryVar::Def *def) {
 }
 } // namespace
 
+bool is_macro_expansion(DB *db, WorkingFile *wf, QueryFile *file,
+                        DocumentSymbol const &ds) noexcept {
+  if (ds.kind == SymbolKind::Macro) {
+    return false;
+  }
+  auto start = ds.range.start;
+  auto syms = findSymbolsAtLocation(wf, file, start, true);
+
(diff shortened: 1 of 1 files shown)


src/messages/textDocument_document.cc
```diff
@@ -130,6 +130,20 @@ template <> bool ignore(const QueryVar::Def *def) {
 }
 } // namespace
 
+bool is_macro_expansion(DB *db, WorkingFile *wf, QueryFile *file,
```
MaskRay, line 133: This can be folded into `ignore` for `QueryVar::Def`

FederAndInk: It's a little bit cumbersome, I'll make a commit, so you can see what it looks like, but we would have to edit all `ignore` functions not just for `QueryVar::Def` because we want to ignore all symbols generated by macro expansion not just variables.

PR #805, opened 2021-06-06
Support semantic highlight, introduced in specification 3.16.0

MaskRay:


codeinred:
Should this be closed now that semantic highlighting has been added, or is there still work to be done?

AkibAzmain:
It works!

Are there any work left?

AkibAzmain:
> It merges successfully and works...

... but not very well.  There are several problems:

- `textDocument/semanticToken/range` doesn't work with multi-line ranges.
- Returns tokens for a old version of the file in question (maybe a version from its cache).
- It declares that `textDocument/semanticToken/full/delta` is supported while initialization, but the method is unusable as `textDocument/semanticToken/full` doesn't give a `resultId`.

MaskRay:
I apologize for the inactivity in development. While my current editor setup has been satisfactory for a few years , I've recently switched to Neovim and have implemented a feature branch supporting semantic tokens to get rainbow semantic highlighting working.

https://github.com/MaskRay/ccls/tree/semantic-tokens (will force push)
neovim-semantic-tokens

MaskRay:
Implemented in master

PR #806, opened 2021-06-07
Move the cache folder to $XDG_CACHE_HOME.

exorcist365:


MaskRay:
This is deliberate to be per-project by default. Using `XDG_CONFIG_HOME` can cause conflicting among projects. Plus, the code as-is does not work on other OSes.

PR #813, opened 2021-07-07
Diagnostic message which starts with a capital letter

EndlessArch:
For the better diagnostic message


MaskRay:
Many command line utilities specifically choose no-capitalization for messages. This goes against the convention and just adds unneeded code complexity.

PR #849, opened 2021-12-01
feat: Add option CLANG_LINK_DYLIB for static build

cathaysia:
Hi, friend.

Clang provides CLANG_LINK_CLANG_DYLIB variable. But it is not a option, it means I can't change it by -DCLANG_LINK_CLANG_DYLIB=OFF.

I had try make CLANG_LINK_CLANG_DYLIB as a options, but policy CMP0077 get warnings, so I create a new option CLANG_LINK_DYLIB to do this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,8 @@ endif()
 
 find_package(Clang REQUIRED)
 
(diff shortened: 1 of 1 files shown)


cathaysia:
sorry, this makes bugs

PR #863, opened 2022-01-23
platform_posix: add support for Haiku

augiedoggie:


MaskRay:
Does Haiku build work with just this change?

augiedoggie:
Yes, it builds cleanly on x86_64 Haiku with llvm/clang 12.0.1. There is a cmake issue with our llvm packages for the other architectures and versions that I haven't gotten around to fixing yet.

PR #873, opened 2022-04-01
Help Clang find bottom of stacks for safety, use desired stack size

dtzWill:
noteBottomOfStack:

Without this, checks against stack space within Clang don't work as
Clang doesn't know where the stack begins.

Needed per-thread, as early as possible.
(on threads using Clang)

Using Clang's desired stack size:

Additionally increase stack size of pthreads to Clang's desired size.

This is presently 8MB, and is used by Clang's stack management
mechanisms to check* if close to stack exhaustion when determining if
there's sufficient space (and warn or run on a new thread with more).
(see runWithSufficientStackSpace)

The constant is available in LLVM 7 onwards.

* (abs(cur - bottom) > DesiredStackSize - threshold)

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -8,6 +8,7 @@
 #include "test.hh"
 #include "working_files.hh"
 
+#include <clang/Basic/Stack.h>
 #include <clang/Basic/Version.h>
 #include <llvm/Support/CommandLine.h>
 #include <llvm/Support/CrashRecoveryContext.h>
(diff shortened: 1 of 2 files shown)


dtzWill:
* Didn't try adding to the windows spawnThread, if that's desired (and otherwise this seems good) I can take a look! :+1: 
* main maybe doesn't need to noteBottomOfStack (if it never calls into clang on that thread), but it doesn't hurt.
* Reviewing the various threads spawned elsewhere (via std::thread) made it seem this (spawnThread) was the best point to add this, but it's maybe not ideal.

This fixes crashes encountered pointing ccls at CIRCT (+llvm submodule), which was my motivation.  I haven't looked into making that reproducible (on other systems), so not sure if that's specific to my setup.  Stack was very deep (just shy of 10k frames) and included many calls to `runWithSufficientStackSpace` that were doing nothing.  For some reason this actually took down my ccls instance (despite CrashRecoveryContext), which made it particularly problematic.

PR #899, opened 2022-09-09
clang format

joshcangit:
Merge commits from @rhcher.


joshcangit:
Sorry, wrong request.
Was testing out multiple other forks since your project was stale.

PR #925, opened 2023-01-01
fix std::optional -> llvm::Optional for up-to-date upstream llvm 16.0.0.

TuringKi:
The upstream llvm seemed using llvm::Optional instead of std::optional for several apis.


MaskRay:
6dc564f2a8608fff778a7d7e8ed0d5acd8d61e6b `Adapt llvm::Optional deprecation` fixed the build.

PR #943, opened 2023-06-28
[SEMA_MANAGER] fix missing PrecompiledPreamble::Build parameter

wafgo:
since llvm commit cc929590ad305f0d068709c7c7999f5fc6118dc9 (CommitDate: Tue Mar 7 08:25:38 2023 -0500) PrecompiledPreamble::Build() has an additional parameter for the default temporary path.

This introduced change breaks the ccls Build. This commit fixes it by checking the CINDEX_VERSION_MAJOR and CINDEX_VERSION_MINOR, and if the version is above 0.63 adds this path to nothing, which bahaves as the previous version according to the code commets.

This is kind of ugly but I don't know a cleaner solution.


MaskRay:
Thank you for the PR. Commit ba38e13b93de1e8787b7e8a22b9a00d35b2b40f2 should be sufficient.

PR #968, opened 2024-08-06
utils: Add `<cstdint>`

thesamesam:
utils uses `uint64_t` without including `<cstdint>` which fails to build w/ GCC 15 after a change in libstdc++ [0]

[0] https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=3a817a4a5a6d94da9127af3be9f84a74e3076ee2

### DIFF
--- a/src/utils.hh
+++ b/src/utils.hh
@@ -6,6 +6,7 @@
 #include <optional>
 #include <string_view>
 
+#include <cstdint>
 #include <iterator>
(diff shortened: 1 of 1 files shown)


thesamesam:
@MaskRay ping

MaskRay:
Thanks:) And thanks for the mention, otherwise I could easily neglect notifications :)

PR #978, opened 2024-11-29
Make sure to get the resource directory from the right clang compiler

rherilier:
The resource directory is always retrieved from the system's default clang compiler instead of the targeted one.

I notice it after removing clang from my system (which is an alias for clang-16) while keeping clang-19 installed.


MaskRay:
I download prebuilt Clang packages on https://releases.llvm.org/download.html then set CMAKE_PREFIX_PATH to the root directory of the extracted tarball. ``Clang_EXECUTABLE` will respect that.

`clang-${Clang_VERSION_MAJOR}` will not work if the distribution only provides `clang` but not something like `clang-19`.

rherilier:
Thanks, I did not expect any differences between Windows and UNICES builds...

PR #988, opened 2025-11-03
resolve the type alias correctly

zhscn:
The struct derived from an alias is missing from type hierarchy before this fix:
```cpp
struct Base {};
struct Derived : Base {};
using BaseAlias = Base;
struct DerivedAlias : BaseAlias {};
```
```
Derive from
Base
└╸Derived
```
The expected output is:
```
Derive from
Base
├╸Derived
└╸DerivedAlias
```


zhscn:
@MaskRay

